PR #453, opened 2014-11-27
Fixes UUID error.

ryansmith94:


ryansmith94:
Spec has changed slightly so this fix is not valid.

Issue #493, opened 2015-01-23
State API returns 400 Bad Request when PUTing a JSON object


garemoko:
**Version**
http://lrs.learninglocker.net/lrs/

**Steps to reproduce the bug**
Using TinCanJS, PUT a document of type application/json twice in a row without using etags. Use the following document contents:

``` {"location":1,"attemptDuration":9083,"attemptComplete":false} ```

**Expected behaviour**
The document should be updated returning 204 No Content. 

**Actual behaviour**
[...]

Issue #491, opened 2015-01-22
State API always(?) returns 412 pre-condition failed when PUTing updates to documents and when Etag headers are used. 


garemoko:
**Version**
Latest from composer as 22nd Jan 2015

**Steps to reproduce the bug**
1. Store an application/json document in the State API. 
2. Get the document in order to get the latest eTag OR get the eTag directly from the database
2. Update the document (PUT) including the If-Match eTag header

**Expected behaviour**
Document should be accepted, returning 204 no content. 

**Actual behaviour**
412 pre-condition failed is returned. 

**Server information**
MongoDB with Basic auth. Running on MAMP. 

**Client information**
OS: OS X 10.10.1
Browser: Chrome

**Additional information**
I've tested the client code I'm using against Wax and Cloud and it works. This also affects the Agent Profile API.

ryansmith94:
Thanks again Andrew, bit of an odd one, I'll take a look next week for 1.3.2.

garemoko:
Couple of notes:

1. Commenting out [this line](https://github.com/LearningLocker/learninglocker/blob/ec7a6ec0245683fa4eb7cf6d57da4a9f760e4dc3/app/locker/repository/Document/EloquentDocumentRepository.php#L322) resolves the 412 issue as the problem seems to be comparing the value I'm sending with quotes with one without quotes, however when I do this I get 400 bad request instead. 

2. Looking at my database I have 2 objects in documentapi. They both have the same _id, the same sha and the same timestamp, which doesn't seem right. It's the state document I'm getting the error when updating. 

```
{
    _id: ObjectId("54c1773c44b444b8020041a7"),
    lrs: "54c1771844b444f0020041a7",
    documentType: "state",
    identId: "http://tincanapi.co.uk/stateapikeys/registrations",
    activityId: "http://id.tincanapi.com/activity/tincan-prototypes/golf-example",
    agent: {
        name: "Andrew Downes",
        mbox: "mailto:andrew.downes@tincanapi.com",
        objectType: "Agent"
    },
    registration: null,
    updated_at: ISODate("2015-01-22T22:18:36Z"),
    sha: "\"DA39A3EE5E6B4B0D3255BFEF95601890AFD80709\"",
    content: {
        "02acbe23-13ee-47ea-a68e-111309ec0d37": {
            created: "2015-01-22T23:18:36+01:00",
            lastlaunched: "2015-01-22T23:18:36+01:00"
        }
    },
    contentType: "application/json",
[...]

Issue #490, opened 2015-01-22
State API requires eTags (should be optional)


garemoko:
**Version**
Latest from composer as 22nd Jan 2015

**Steps to reproduce the bug**
1. Store an application/json document in the State API. 
2. Update the document (PUT) omitting the etag header. 

**Expected behaviour**
Document should be accepted, returning 204 no content. Etags are not required for the State API, only the Agent and Activity Profile APIs. 

**Actual behaviour**
400 Bad request is returned. 

**Server information**
MongoDB with Basic auth. Running on MAMP. 

**Client information**
OS: OS X 10.10.1
Browser: Chrome

**Additional information**
I've tested the client code I'm using against Wax and Cloud and it works.

ryansmith94:
Thanks for the issue and following the guidelines Andrew :+1: this is part of the spec I'm aware of but I was unaware that we didn't do that. This needs to be reported as an issue on the xAPI test suite too I think.

ryansmith94:
I'll try to take a look at fixing this next week for 1.3.2.

ryansmith94:
Hey @garemoko please can you pull [issue/490](https://github.com/LearningLocker/learninglocker/tree/issue/490) and tell me if that fixes the issue for you? Thanks :smile:

davidpesce:
After removing etag from plugin, I'm getting the error, "Error attempting to set registration data to State API."

garemoko:
what's the full error test below that message? What error code is returned by the LRS?

ryansmith94:
From what I can tell the header is not required in the 490 branch and the ETag isn't checked when storing a document via the State API either. I must be missing something. I'll run the conformance tests against the branch and see how that goes.

davidpesce:
Full error output:
```
The Learning Record Store is not available. Please contact a system administrator.

Error attempting to set registration data to State API.

array(3) {
  ["contents"]=>
  bool(false)
  ["metadata"]=>
  NULL
  ["code"]=>
  array(12) {
    [0]=>
    string(22) "HTTP/1.0 404 Not Found"
    [1]=>
    string(35) "Date: Wed, 18 Feb 2015 13:17:29 GMT"
    [2]=>
    string(39) "Server: Apache/2.4.6 (CentOS) PHP/5.6.4"
    [3]=>
    string(19) "Vary: Authorization"
    [4]=>
    string(23) "X-Powered-By: PHP/5.6.4"
    [5]=>
[...]

PR #528, opened 2015-02-26
Passes conformance tests.

ryansmith94:
#490 #491 #493

### DIFF
--- a/app/controllers/xapi/DocumentController.php
+++ b/app/controllers/xapi/DocumentController.php
@@ -95,7 +95,7 @@ abstract class DocumentController extends BaseController {
 
     if ($document) {
       return \Response::json(null, BaseController::NO_CONTENT, [
-        'ETag' => $document->sha
+        'ETag' => '"'.$document->sha.'"'
       ]);
     } else {
       throw new \Exception('Could not store Document.');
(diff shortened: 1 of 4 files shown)


app/controllers/xapi/DocumentController.php
```diff
@@ -83,6 +83,8 @@ public function store() {
     $data['content_info'] = $this->getAttachedContent('content');
     $data['ifMatch'] = \LockerRequest::header('If-Match');
     $data['ifNoneMatch'] = \LockerRequest::header('If-None-Match');
+    $data['ifMatch'] = isset($data['ifMatch']) ? trim($data['ifMatch'], '"') : null;
+    $data['ifNoneMatch'] = isset($data['ifNoneMatch']) ? trim($data['ifNoneMatch'], '"') : null;
```
fugu13: you can't just trim these; that'll result in accepting illegal values (with extra quotes). You could either make sure to just trim the outer quotes, or a cleaner approach would be to add the quotes to the hash to make the document's ETag, then compare the ETag directly with the headers in question.

ryansmith94: I'm aware of this already. Once the ETags are wrapped in quotes for the xAPI test suite I'll change this. For now though, quotes can't appear in the hash, so despite trimming additional quotes the world will probably spin on.

PR #523, opened 2015-02-19
Add a new line for next chunk.

kzkamiya:
When I tried csv exporting function of Learning Locker, a csv file of over 1000 rows did't include a newline in each 1000 rows.
So, I added a newline for next chunk.

PR #536, opened 2015-03-04
Fixes newline bug.

ryansmith94:
This behaviour was changed in #523. This PR reverts the decoding so that new lines do not cause issues.

### DIFF
--- a/app/controllers/api/Exports.php
+++ b/app/controllers/api/Exports.php
@@ -109,7 +109,7 @@ class Exports extends Resources {
 
             foreach ($keys as $key) {
               // Decode unicode characters
-              array_push($values, json_decode('[' . $statement[$key] . ']', true)[0]);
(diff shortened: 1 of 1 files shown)


ryansmith94:
Sorry if decoding was important to you @kzkamiya. Unfortunately if a new line is contained within the value it screws up the CSV. You're welcome to do some more work around this. For example we could decode if we're certain it contains no new lines or quotes (or anything else that might screw up the CSV).

kzkamiya:
Hi, @ryansmith94. I see. My changed code has some problems in the CSV. I will think another idea for decoding.

ryansmith94:
Hey @kzkamiya, yeah basically we can't decode new lines or anything that is part of the CSV syntax (in our case quotes, commas, and new lines). Awesome, that would be good thanks :+1:

PR #617, opened 2015-05-05
Adds OAuth support.

ryansmith94:
Any credentials previously used for Basic Auth can now be used to request access tokens via OAuth. Please follow the steps below to use OAuth.

1. Request access token via `POST www.example.com/oauth/access_token` with `client_id` set to your key/username, `client_secret` set to your secret/password, and `grant_type` set to "client_credentials".
2. Receive `token` in response from step 1.
3. Make requests to the API by setting the `Authorization` header to "Bearer" followed by a space and the `token` you received in step 2.

**Notes**
The code inside the Helpers for this needs to be moved to the ClientRepository really. However, the LRS credentials need to be removed otherwise this code will be duplicated in the LrsRepository. We should remove the credentials from the Lrs model and just use credentials from the Client model.


ht2:
@ryansmith94 Is this URL documented anywhere other than this PR?

ryansmith94:
I don't think it is :-\ @ht2

ht2:
No prob, its actually clearly used here: https://github.com/LearningLocker/learninglocker/blob/develop/app/tests/routes/OAuthTest.php

ryansmith94:
Ah good spot :-)

ProdEng:
This link seems to be no longer valid, where might an example now be located?

ht2:
@ProdEng https://github.com/LearningLocker/learninglocker/blob/v1/app/tests/routes/OAuthTest.php

ht2:
@ProdEng Please note that is V1 only

ProdEng:
Thank you sir, any OAuth support within V2?

andrewhickey:
Not yet, but it's on the roadmap

ProdEng:
Ta, possible time frame for production release?

ht2:
@ProdEng There are no timeframes. If you would like to contribute and start some work on that feature then contributions are always appreciated.

ProdEng:
Love to, but not a developer just a lowly CIO/CTO.  Keep up the great work though as I am riding your horse to market.

Best,

Jon

juangon:
Hi  @ryansmith94 , been trying to use /oauth2/token endpoint with client_id etc without luck. How these parameters should be sent? As JSON? As URL parameters? Thanks!

PR #620, opened 2015-05-05
Adds scopes to Clients.

### LINKED ISSUE #121: Moodle integration - security/authorisation
I've been thinking about Moodle/LRS integration for the purposes of security when launching SCORM-style JavaScript based e-learning. 

## High level process
I'd like to implement something along the lines of the following:

1. The Moodle authenticates with the LRS server side via ~~OAuth~~ Basic HTTP. It has special permissions on the LRS to do the rest of the process. 

2. Every time somebody launches an experience, the Moodle requests a new set of Basic credentials for that launch. The Moodle *requests* certain permissions associated e.g. only able to issue statements about a specific actor an/or activity, only valid for a certain period of time or time gap between statements (it expires at 4:30pm or 30 minutes after the last statement received, for example). 

3. When receiving statements (and documents) the LRS check against the permissions relating to the credentials and accepts or rejects accordingly.

To achieve this we need to:
### Design
* Define a set of permissions
* Define ~~an API endpoint and~~ (done) data structure for requesting and returning basic credentials
* Discuss with the wider community, particularly CMI5 and the XAPI Working Group

### Moodle - all done
https://github.com/garemoko/moodle-mod_tincanlaunch
* ~~Implement the mechanism for requesting credentials as design above~~ 
* ~~Make this configurable~~ 

### Learning Locker
* ~~Complete OAuth~~ Not required
* ~~Enable multiple sets of Basic credentials per LRS~~ Done
* Create a permissions management system and associate permissions with credentials as designed above
* Expose the permissions system via API as designed above -API done, needs permissions elements.

ryansmith94:
Uses "scopes" (permissions) which should resolve #121.

### DIFF
--- a/app/controllers/ClientController.php
+++ b/app/controllers/ClientController.php
@@ -48,10 +48,12 @@ class ClientController extends BaseController {
     $lrs = $this->lrs->show($lrs_id, $opts);
     $lrs_list = $this->lrs->index($opts); 
 	  $client = $this->client->find($id);
+    $scopes = \App::make('db')->getMongoDb()->oauth_scopes->find()->getNext()['supported_scopes'];
 	 	return View::make('partials.client.edit', [
       'client' => $client,
       'lrs' => $lrs,
-      'list' => $lrs_list
+      'list' => $lrs_list,
+      'scopes' => $scopes
 		]);
   }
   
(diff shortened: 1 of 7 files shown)


ryansmith94:
Closed in favour of #645

PR #632, opened 2015-05-13
Fixes timestamps.

ryansmith94:
adlnet/xAPI-Spec#541


ryansmith94:
This isn't a problem at the moment because the Statement Factory corrects the timestamp.

Issue #609, opened 2015-04-27
Filtering by actor's account does not work.


ryansmith94:
**Version**
1.3.10

**Steps to reproduce the bug**
1. Insert a statement with an `actor` identified by an `account`, where the `homePage` is "http://www.example.com" and the `name` is "1".
2. Attempt to retrieve that statement with the request below.
```
GET /data/xAPI/statements?agent={"account":{"homePage":"http://www.example.com/","name":"1"}} HTTP/1.1
Host: www.example.com
Authorization: Basic YOUR_AUTH
X-Experience-API-Version: 1.0.1
```

**Expected behaviour**
Step 2 should return the inserted statement in step 1 because of the xAPI specification (see additional information)

**Actual behaviour**
Step 2 should returns nothing.

**Server information**
MongoDB with Basic Auth

**Additional information**
[xAPI GET specification](https://github.com/adlnet/xAPI-Spec/blob/master/xAPI.md#723-getstatements)
[xAPI account specification](https://github.com/adlnet/xAPI-Spec/blob/master/xAPI.md#4124-account-object)

PR #634, opened 2015-05-13
adding multiple "or" actor queries as well as being able to do reports on actors via account/mbox/openid/mbox_sha1sum

loongchan:
Hey Ryan,
This should add the ability to create reports on actors that do not use mbox, but other things as well.  Also, multiple actors can be used to create "or"'ed queries instead of default "and" queries, which I don't believe make as much sense in this case.

Best,
Loong

### DIFF
--- a/app/locker/repository/Query/EloquentQueryRepository.php
+++ b/app/locker/repository/Query/EloquentQueryRepository.php
@@ -31,7 +31,7 @@ class EloquentQueryRepository implements QueryRepository {
                                     $subVal_array = get_object_vars($subVal);
                                     $query->orWhere(function($query) use ($subVal_array, $value) {
                                         foreach ($subVal_array as $key => $val) {
-                                            $auery->where($value[0] . '.' . $key, '=', $val);
+                                            $query->where($value[0] . '.' . $key, '=', $val);
                                         }
                                     });
                                 } else {
--- a/app/locker/repository/Query/EloquentQueryRepository.php
+++ b/app/locker/repository/Query/EloquentQueryRepository.php
@@ -27,7 +27,16 @@ class EloquentQueryRepository implements QueryRepository {
                     $statements->where(function($query) use ($filter) {
                         foreach ($filter[2] as $value) {
                             foreach ($value[1] as $subVal) {
-                                $query->orWhere($value[0], '=', $subVal);
+                                if (is_object($subVal)) {
+                                    $subVal_array = get_object_vars($subVal);
+                                    $query->orWhere(function($query) use ($subVal_array, $value) {
+                                        foreach ($subVal_array as $key => $val) {
+                                            $auery->where($value[0] . '.' . $key, '=', $val);
+                                        }
+                                    });
+                                } else {
+                                    $query->orWhere($value[0], '=', $subVal);
+                                }
                             }
                         }
                     });


ryansmith94:
This doesn't appear to have been merged correctly. See https://github.com/LearningLocker/learninglocker/commit/66f0152981ad86812c5b96dff67000b4db311afc for example.

ryansmith94:
Something isn't right with this.

**Version**
ubc/learninglocker at fcb4637 on version 1.4.0.

**Steps to reproduce**
1. Insert one statement with `actor.mbox_sha1sum` set to "7a8744f315a8d5a5b14b1af4b524912952059d3b".
2. Insert one statement with `actor.account.homePage` set to "http://www.example.com/extended" and `actor.account.name` set to "1".
3. Insert one statement with `actor.account.homePage` set to "http://www.example.com/extended" and `actor.account.name` set to "2".
4. Create a report.
5. Add the following actor "mbox_sha1sum:7a8744f315a8d5a5b14b1af4b524912952059d3b".
6. Add the following actor "account:http://www.example.com/extended / 1".
7. Add the following actor "account:http://www.example.com/extended / 2".
8. Click "Save".
9. Click "Statements".

**Expected behaviour**
Step 9 should show three statements from steps 1 to 3 because I sent one statement for each of the three actors added in steps 5 to 7.

**Actual behaviour**
Step 9 actually shows one statement (sent in step 1) for the first the actor (added in step 5).

**Additional information**
image
*Actors in my report. Note that it displays them in a different order when revisiting the report.*

image
*Statements shown in step 9*

ryansmith94:
This issue appears to be with accounts. I tried to filter with just one actor whose IFI was an account and it didn't work.

I imagine it's because your probably trying to filter the `actor.account` property rather than filtering the `actor.account.name` and `actor.account.homePage`. We used to have this problem in the xAPI for statements (#609).

loongchan:
Hey Ryan, 
I tried to reproduce the bug (great bug report btw!)
statements I sent:
screen shot 2015-05-15 at 9 34 18 am
screen shot 2015-05-15 at 10 00 50 am
screen shot 2015-05-15 at 10 01 15 am
screen shot 2015-05-15 at 10 01 31 am


query:
screen shot 2015-05-15 at 9 36 49 am

results of report:
screen shot 2015-05-15 at 9 37 38 am

aka, I can't reproduce. I WANT to reproduce the bug.  
Also, regarding the ordering.  yeah, noticed that.  I was aiming for functional first.  I'll look into it.  

Thanks for looking into this!

ryansmith94:
Hey @loongchan, sorry for the delay. Thanks, I just used our guidelines, I use them pretty much everywhere now.

I've just cleared my LRS and gone through those steps again and it does reproduce it for me.

No worries about the ordering to be honest, I just wondered if that might have an effect.

No problem. Can you confirm what commit you're currently at and the code within the `where` method inside "app/locker/repository/Query/EloquentQueryRepository.php"?

loongchan:
Hey Ryan,
I just did a fresh install from cloning the https://github.com/ubc/learninglocker repo, running composer, creating a db, setting up configuration stuff, generating statements, report... seems to work still. 

May I ask what the guidelines are?  I'll follow those steps to reproduce the bug.  I want to go back to using your repository!  :smiley: 

With regards to `where` method inside "app/locker/repository/Query/EloquentQueryRepository.php", it's just the same as the repo.  

```php
public function where($lrsId, array $filters) {
    $statements = \Statement::where(self::LRS_ID_KEY, $lrsId);
    foreach ($filters as $filter) {
      switch ($filter[1]) {
        case 'in': $statements->whereIn($filter[0], $filter[2]); break;
        case 'between': $statements->whereBetween($filter[0], [$filter[2], $filter[3]]); break;
        case 'or':
                if (!empty($filter[2]) && is_array($filter[2])) {
                    $statements->where(function($query) use ($filter) {
                        foreach ($filter[2] as $value) {
                            foreach ($value[1] as $subVal) {
                                $query->orWhere($value[0], '=', $subVal);
                            }
                        }
                    });
                }
                break;
        default: $statements->where($filter[0], $filter[1], $filter[2]);
      }
    }
    return $statements;
  }
```

ryansmith94:
That's so odd, I might try doing a fresh install again.

Well we don't actually document the guidelines any more because people can use the templates in the [contribution guide](https://github.com/LearningLocker/learninglocker/blob/develop/contributing.md). However, I do have [my own guidelines stored as a gist](https://gist.github.com/ryansmith94/d600afe7b267913c994d) for all my repositories external to HT2.

Thanks for that snippet, you `where` method looks identical to mine so that seems ok.

ryansmith94:
This is pretty frustrating, I'm really sorry about the delay, but I've just done a fresh install, followed my steps to reproduce the issue and it still exists.

ryansmith94:
The issue still exists if I just follow steps 3, 4, 7, 8, and 9 (so just filtering by account).

ryansmith94:
Hey @loongchan, so I've figured out the issue here. The issue is that I'm inserting my accounts with the `homePage` before the `name` (see snippet 1 below) and you must be inserting the accounts with the `name` before the `homePage` (see snippet 2 below) in steps 1-3.

```json
"account": {
  "homePage": "http://www.example.com/extended",
  "name": "1"
}
```
*Snippet 1 - What I insert.*

```json
"account": {
  "name": "1",
  "homePage": "http://www.example.com/extended"
}
```
*Snippet 2 - What you insert.*

As I mentioned in [an earlier comment](#issuecomment-102345861) this is caused by matching on the entire object (`statement.actor.account`) instead of the individual keys (`statement.actor.account.name` and `statement.actor.account.homePage`).

loongchan:
Hey Ryan,
I fixed that in 0fed96e18fd2c3c25ee31e3d4cbca887519e1364 .  I matched the entire object because you made a great point about needing both homePage and name because there could be more than one statement.actor.account.name.  
Also, I'm not too well versed in Mongodb, so it took me quite a bit of time to figure out order DOES matter.

ryansmith94:
Hey @loongchan,

Unfortunately that doesn't fix it, because the specification doesn't define that the `name` needs to come before the `homePage` or vice-versa hence it could be inserted either way. You'll need to match on all of the properties (name, homePage) within the object not entire the object.

ryansmith94:
Thanks for all your effort on this by the way. After this is merged you'll be the largest contributor to Learning Locker outside of HT2.

loongchan:
Once again, good points about name and homePage.  I agree with you I thought it was funny when I found the error.  
It seems like if you pass an object to Mongo that it will match it only if order also matches it.  I will look into making it do "and" query with name and homePage for account queries.... though it's already in an "or" query.  
Tomorrow!  :smile:

loongchan:
Hey Ryan,
I updated the query portion so that it will do "and" queries with name and homePage for account.

Best,
Loong

ryansmith94:
Going to give this a test now, hopefully it's all good this time, really want to merge it! :smile:

ryansmith94:
Damn it, the statements list correctly displays three now (nice work), but the graph only shows one.

ryansmith94:
I have a fix for that issue, so I'm going to merge it and apply the fix :+1:

ryansmith94:
It's in! Thanks for all of your hard work! Congratulations on becoming the largest contributor to Learning Locker outside of HT2! :+1:

PR #642, opened 2015-05-22
Changes related to Username and Password wrapping in client partials.

hallshouse:
Changed table layout options in manage.blade.php as well as line length in item.blade.php.

### DIFF
--- a/app/views/partials/client/item.blade.php
+++ b/app/views/partials/client/item.blade.php
@@ -1,7 +1,7 @@
 <td>{{ $client->authority['name'] }}</td>
-<td>{{ chunk_split($client->api['basic_key'], 40, '') }}</td>
-<td>{{ chunk_split($client->api['basic_secret'], 40, '') }}</td>
-<td><center>
+<td>{{ $client->api['basic_key'] }}</td>
+<td>{{ $client->api['basic_secret'] }}</td>
+<td class="ico_col">
   <a
     href="{{ URL() }}/lrs/{{ $lrs->_id }}/client/{{ $client->_id }}/edit"
     class="btn btn-info btn-sm"
@@ -9,5 +9,5 @@
   >
     <i class="icon-pencil"></i>
   </a>
-</center></td>
-<td><center>@include('partials.client.forms.delete')</center></td>
+</td>
+<td class="ico_col">@include('partials.client.forms.delete')</td>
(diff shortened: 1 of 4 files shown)


app/views/partials/client/item.blade.php
```diff
@@ -1,7 +1,7 @@
 <td>{{ $client->authority['name'] }}</td>
-<td>{{ chunk_split($client->api['basic_key'], 20, '</br>') }}</td>
-<td>{{ chunk_split($client->api['basic_secret'], 20, '</br>') }}</td>
-<td>
+<td>{{ chunk_split($client->api['basic_key'], 40, '') }}</td>
```
ryansmith94: The `chunk_split` isn't required if you're just displaying it one line.

app/views/partials/client/item.blade.php
```diff
@@ -1,7 +1,7 @@
...
+<td>{{ chunk_split($client->api['basic_secret'], 40, '') }}</td>
+<td><center>
```
ryansmith94: I think the `center` tag is frowned upon because it's more of a styling thing so it belongs in the CSS.

app/views/partials/client/manage.blade.php
```diff
@@ -29,8 +29,8 @@
             <th>{{trans('site.name')}}</th>
             <th>{{trans('site.username')}}</th>
             <th>{{trans('site.password')}}</th>
-            <th>{{trans('site.edit')}}</th>
-            <th>{{trans('site.delete')}}</th>
+            <th width="5%"><center>{{trans('site.edit')}}</center></th>
```
ryansmith94: [Again](#r30899153) `center` tag is frowned upon and so is the width property here.

app/views/partials/client/item.blade.php
hallshouse: Glad you picked these up.
Now that the changes are targeted for 1.4.3, I'll be happy to make all the adjustments required to do these changes properly. I'll update my changes to use CSS where required, and investigate the removal of chunck-split and test before submitting again.

public/assets/css/app.css
```diff
@@ -215,6 +215,10 @@ table.break-words td, table.break-words th {
   word-wrap: break-word;
 }
 
+.ico_col {
```
ryansmith94: Sorry to be a pain but can we remove this and use the `text-center` class provided by [bootstrap](http://getbootstrap.com/css/) (this a dependency of ours already so we might as well use it). I should have mentioned this before sorry.

ryansmith94:
Thanks @hallshouse :+1:

Issue #695, opened 2015-07-16
Reporting with "Scaled" filters doesn't work


flipace:
**Version**
develop at d588864dc9ea0ac464823f47ecc5c3a0d039b5cb on 1.6.1

**Steps to reproduce the bug**
Create a new report, use one of the "range" filters for "With what?" (e.g. raw)

**Expected behaviour**
Get all statements where the raw score is between the set values.

**Actual behaviour**
No statements are returned because query is not valid.

It should be something like
{"statement.result.score.raw":{"$gte":0, "$lte":10000}}
but it is
{"statement.result.score.raw":{"$in":["<>", 0, 10000]}}

**Client information**
OS: OSX 10.10.1

daniel-abbey:
@flipace Many thanks for reporting and resolving this issue.

PR #698, opened 2015-07-17
Fixed filtering of statements with "between" values

flipace:
This fixes issue #695


daniel-abbey:
@flipace Many thanks for spotting and resolving this issue.

PR #725, opened 2015-09-02
Fixes 'Verified' button styles on Admin Dashboard

daniel-abbey:


daniel-abbey:
But it has no event bound to it. It is called from within `verifyUser()`  @ryansmith94

daniel-abbey:
@ryansmith94 I removed the function and moved its contents to the AJAX success function

PR #727, opened 2015-09-04
Adds interval and time graph to reports.

ryansmith94:
I would add it to the admin dashboard but it only works with the aggregation API at the moment (which is limited to LRSs).


ryansmith94:
Not ready to merge due to failing Travis builds and because I want to add it to the LRS dashboards.

ryansmith94:
Fixed tests and added graphs to LRS dashboards. To add to admin dashboard, a new route will need to be created for aggregation over all LRSs and the public/assets/js/libs/exp/src/AggregationRepo.ts file will need to be changed to allow for other aggregation routes.

ryansmith94:
@daniel-abbey this needs to be demonstrated to James, Andrew, and Ben next Monday before merging, so it'll have to wait until the next release :+1: They may want to tweak some parts of it before it's merged.

daniel-abbey:
Thanks Ryan

ryansmith94:
No worries :-)

PR #729, opened 2015-09-04
Fixes state API.

ryansmith94:


ryansmith94:
Not so sure about this one. I was going through old branches trying to remove them and this appears to be a legit fix. But I can't think why I wouldn't have merged it so just keep a look out for any issues with this one.

PR #726, opened 2015-09-03
Fixes LRS statement counts on Admin Dashboard

daniel-abbey:

PR #730, opened 2015-09-04
Attempts to fix Lrs count performance.

ryansmith94:
Needs testing. Best to test with Andrew or the Staging server since they have the most "non-production" data. Not sure if this effects #726.


ryansmith94:
Shouldn't be merged without testing if it does actually improve performance. If it doesn't then it should be closed and the branch should be deleted.

daniel-abbey:
No conflicts with #726 as far as I can see

ht2:
Helps address issues found in #760

PR #745, opened 2015-09-22
Combined branches for MongoId issues and adding Rackspace cloud storage option

daniel-abbey:
Contains a merging of the branches `issue/lrsuserids`, `issue/mongo-object-role` and `issue/rackspace-url-type`.


daniel-abbey:
Merged into #748.

PR #748, opened 2015-09-24
Combined Pondermatic and other branches

daniel-abbey:
Due to some overlapping functionality of a handful of branches, we've re-merged into one "super" branch! We'll handle the PR through this, but for the most part @pondermatic this is just your original PR plus some extra bits we noticed around the place. It also combines a new way to pull Rackspace Cloud Files.


ht2:
Great work on all this @pondermatic, we owe you a drink (or 5)! Just some info, we had already made some similar changes in other branches (I believe you'd seen them and left a couple of comments), so I thought best we merge all of them together - the main one was the migration, which we had actually deployed in some test environments. To avoid problems going forward I've left in the "defunct" migration, with reference in it to your version that handles everything correctly (and in one place). I just spent a bit of time with @daniel-abbey going through all of these and ensuring everything looks ok - hopefully this all passes and we can get it all merged in!

pondermatic:
Commit d8fe43238244b3ac5a973eab131bb204c1d5254b is not needed. Controllers should not convert a model's data type. The statement's lrs_id conversion is handled by `app/locker/repository/Statement/EloquentReader->where()`.

ht2:
Your absolutely right, we hadn't noticed the addition to the EloquentReader (and it will still work as it would cast it to string before making it a new MongoId!) We'll make an amend and rerun it

ht2:
Good work everybody!

Issue #758, opened 2015-11-30
getMatch does not use indexed `lrs_id`


ht2:
The `getMatch` method in `app/locker/data/BaseData.php` does not use the `lrs_id`. This could lead to slower queries on a non-indexed and string only ID field.

ht2:
I believe the fix is to simply change to...

```
  protected function getMatch( $lrs ){
    return array('lrs_id' => new \MongoId($lrs));
  }
```

I will submit a PR shortly.

ht2:
Merged into develop

PR #759, opened 2015-11-30
Change the getMatch method to use the lrs_id field with MongoId

ht2:
Addresses issue #758


ht2:
OK... looks like pear.php.net is currently unresolvable.

Issue #760, opened 2015-12-01
Slow counts on large collections


ht2:
It seems that the `count()` method through the jenssegers/laravel-mongodb package is not using the native count method available on collections - I believe it might even be trying to get the full set of documents for a matching query, and then count them server side (which is ridiculous). This is mainly causing on the dashboard where we run some intensive counts.

I have a fix I am currently testing and will push it up to a branch shortly.

ryansmith94:
Not sure if this is in anyway related to what I was working on in #730.

ryansmith94:
Current way to count statements is to find all the LRS IDs, then loop over the statements for each LRS ID finding statements with that LRS ID. Basically you loop over all the statements for each LRS. #730 only loops over the statements once.

ht2:
@ryansmith94 we rewrote the Dashboard aggregation queries in a similar fashion and that is all pushed to `develop` - we utilised the new aggregate features in Mongo 2.6 to get it done in one query. The problem in this issue is simply the `->count()` method from the Eloquent Query builder which just doesn't use the `MongoCollection` native count method (http://php.net/manual/en/mongocollection.count.php)

ht2:
OK, this commit should do what we need: dbec5dac3205d310ebdc7833c5bbad64268073a2

ht2:
Merged in Ryan's fix for the statement count on the LRS list page too (which uses the native method as well)

ryansmith94:
Awesome :+1:

ht2:
I've unified the methods into one repo above - the dashboards and LRS list now call from one repo method

ht2:
So, after further investigation we believe this may be down to the stored value being queried and indexed as a string. As such we have a pull request incoming that adds a new MongoDate value to the top of the statement's document. It provides migrations to do this for existing data, and ensures all new data also is inserted with these fields. It also updates any referenced statements through the migration as well as a handful of indexes (added in the background). All of this allows for much quicker sorting on statements as the "statement.stored" was the default sorting mechanism, and often the default field for querying on time (e.g. since/until in the xAPI/statements/GET).

PR #761, opened 2015-12-01
Changed statement count method to use native count on BaseDashboard

ht2:


ht2:
Addresses issues in #760

PR #762, opened 2015-12-01
Issue/lrs statement counts

ht2:
Unifies the method for getting statement counts using the native count method. Addresses issue #760

### DIFF
--- a/app/controllers/SiteController.php
+++ b/app/controllers/SiteController.php
@@ -121,7 +121,6 @@ class SiteController extends BaseController {
     $opts = ['user' => \Auth::user()];
     $lrss = $this->lrs->index($opts);
     $lrs_repo = $this->lrs;
-    //$counts = $this->lrs->getCounts();
 
     $collection = \DB::getCollection('statements');
 
--- a/app/controllers/SiteController.php
+++ b/app/controllers/SiteController.php
@@ -120,11 +120,14 @@ class SiteController extends BaseController {
   public function lrs(){
     $opts = ['user' => \Auth::user()];
     $lrss = $this->lrs->index($opts);
-    $counts = $this->lrs->getCounts();
+    $lrs_repo = $this->lrs;
+    //$counts = $this->lrs->getCounts();
 
+    $collection = \DB::getCollection('statements');
 
-    return Response::json(array_map(function ($lrs) use ($counts) {
-      $lrs->statement_total = isset($counts[$lrs->_id]) ? $counts[$lrs->_id] : 0;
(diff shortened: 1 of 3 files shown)


app/controllers/SiteController.php
```diff
@@ -120,11 +120,13 @@ public function getGraphData(){
...
     $lrss = $this->lrs->index($opts);
-    $counts = $this->lrs->getCounts();
+    $lrs_repo = $this->lrs;
 
+    $collection = \DB::getCollection('statements');
 
-    return Response::json(array_map(function ($lrs) use ($counts) {
-      $lrs->statement_total = isset($counts[$lrs->_id]) ? $counts[$lrs->_id] : 0;
+    return Response::json(array_map(function ($lrs) use ($lrs_repo) {      
+      $query = ['lrs_id'=>new \MongoId($lrs->_id)];
```
ryansmith94: I think line 128 can be removed. I can't see where it's used.

app/locker/data/dashboards/BaseDashboard.php
```diff
@@ -34,17 +37,8 @@ public function getGraphData(\DateTime $startDate = null, \DateTime $endDate = n
...
-
-    $query = [];
-    
-    //Limit by LRS
-    if( $this->has_lrs ){
-      $query['lrs_id'] = new \MongoId($this->lrs);
-    }
-
-    return $collection->count($query);
+    $lrs_id = $this->has_lrs ? $this->lrs : null;
```
ryansmith94: If there's a method that *needs* to be called externally before this method to set `$this->has_lrs` can that be commented somewhere?

ht2: The constructor sets it as a boolean depending on whether an $lrs (id) was passed in as an argument

app/controllers/SiteController.php
ht2: Yea you're right, I moved the count to the repo method and the query didn't get removed

app/locker/data/dashboards/BaseDashboard.php
ht2: Sorry that wasn't quite correct, it is set in the constructor, but explicitly depending on the type of dashboard that extends the BaseDashboard (which should really be abstract). E.g the LrsDashboard declares itself as having an LRS...
https://github.com/LearningLocker/learninglocker/blob/issue/lrs-statement-counts/app/locker/data/dashboards/LrsDashboard.php#L11

ryansmith94: Ahhhh I get it. Probably could be static then or $this->lrs could be set to null and remove the variable.

PR #765, opened 2015-12-16
Making composer.json a bit more usefull.

WillSkates:
I've been trying to run the test suite in our environment and these changes helped.

I'm not sure whether "git@github.com:LearningLocker/oauth2-server-laravel" is actually used at this point as I can't find any reference to it in composer.lock. I may just be missing something obvious though.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ services:
   - mongodb
 
 install:
+  - pecl install mongo
   - php -r "readfile('https://getcomposer.org/installer');" | php
   - php composer.phar install --no-interaction --dev
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ services:
   - mongodb
 
 install:
-  - pecl install mongo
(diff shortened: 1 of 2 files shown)


WillSkates:
Seems that the legacy mongo driver is no longer in the provided by default by Travis. I've added a fix for that as well.

.travis.yml
```diff
@@ -7,6 +7,7 @@ services:
   - mongodb
 
 install:
+  - echo y | pecl install mongo
```
ryansmith94: This isn't needed here.

WillSkates: https://travis-ci.org/LearningLocker/learninglocker/builds/97171246

WillSkates: I agree it shouldn't but for some reason it is.

ryansmith94: Pasting in extracts of the GItter chat for future reference.

> two mongo extensions for php now
https://pecl.php.net/package/mongo
https://pecl.php.net/package/mongodb

PR #766, opened 2015-12-23
Issue/stored date

ht2:
Addresses issues with the "stored" value being queried and indexed as a string. 

This PR adds a new MongoDate value to the top of the statement's document. 

It provides migrations to do this for existing data, and ensures all new data also is inserted with these fields. 

It also updates any referenced statements through the migration as well as a handful of indexes (added in the background). 

All of this allows for much quicker sorting on statements as the "statement.stored" was the default sorting mechanism, and often the default field for querying on time (e.g. since/until in the xAPI/statements/GET).

Probably addresses some speed concerns seen in #760


ryansmith94:
Sounds and looks good to me :+1:

ryansmith94:
I'm sure you're already aware, but it fails the conformance tests.

ryansmith94:
This is still failing conformance tests @ht2

ht2:
Sorry, thought the build said it was OK on both accounts

ht2:
Actually are you sure? the Mongodates branch that was merged in fixed that issue. It was failing due to the microseconds not being set properly on MongoDates, and @andrewhickey and I handled that here: https://github.com/LearningLocker/learninglocker/commit/012e5844831412ddaa96e105d09f2118506a5bbf

ryansmith94:
Hmmm well the last run of the conformance tests on this PR definitely failed. I'm trying to change the test script on Travis so that it produces an exit code, but it's difficult with Grunt.

ht2:
Perhaps worth me opening up a PR from develop -> master to get all these merged branches running a PR so we can have a good look?

ryansmith94:
That won't help. The conformance tests only run when merging into develop to improve the build speed.

ht2:
Gotcha, ok well for the mo lets mothball this and move to #784 to try and rectify. `develop` is billed as being "unstable" anyway, and I can't replicate the conformance tests problem anymore - so I'm happy to say that whilst `develop` isn't truly "conformant" its better than having the slow count issues and mongoDate problems. I believe the only issue (if the tests _are_ still correct) is that querying with microseconds might not catch everything.

ryansmith94:
The problem is that the tests bail on the first error, so there may be more problems that we're unaware of. So I don't think we can really leave it. If there is a problem with the conformance tests it needs to be fixed, however, I suspect that is not the case here.

ryansmith94:
Now that I've merged #784 we should be able to avoid accidental merges when the conformance tests fail. How can we fix this issue with the conformance tests? You said you were longer reproducing the error locally?

ht2:
So the issue we found was that because timestamps weren't being stored with microtime, the conformance test was making really specific calls on the since & until's (e.g. it would check for statements between 10:55:44.226900+00:00 and 10:55:44.444000+00:00). Because the way we were making MongoDates before wasn't passing in the microtime, it wouldn't return back correct amount of statements as it was simply rounding down to the nearest second.

We changed that here: https://github.com/LearningLocker/learninglocker/commit/012e5844831412ddaa96e105d09f2118506a5bbf#diff-3f849cce6badabe5ee5d38042a1a9ec7R97 so the only thing I can think is there is _other_ places that need that amendment still. I'll have a look into the GET xapi/statements query and see what it s querying on. Is it worth making a new issue and branch for fixing this conformance stuff in the meantime?

ryansmith94:
I think we should definitely create a new branch and PR that focuses on the one failing test.

ryansmith94:
I'm just looking to see if we updated the StatementIndexer or not.

ryansmith94:
Ok so it's very likely that the problem is on line 48 and 58 in the [StatementIndexer](https://github.com/LearningLocker/learninglocker/blob/develop/app/locker/repository/Statement/EloquentIndexer.php#L48).

ht2:
Good shout, it should _never_ be querying the string timestamp value in the `statement` anymore (`statement.timestamp`)

ryansmith94:
It looks like it is querying the right key (`stored`) but constructing the date correctly.

ht2:
Yep, ok - I'll get a new branch off develop and make the relevant amends. Want to create an issue to reference or is this fine?

ryansmith94:
I think this is fine, since it was a problem caused by this PR that hasn't been released.

ht2:
https://github.com/LearningLocker/learninglocker/blob/issue/mongodate/app/locker/repository/Statement/EloquentIndexer.php

OK, so I still reckon there are a few places that might not do the microtime properly (I'm looking at you StatementRefs)

ryansmith94:
Ok, in terms of querying the statements, that should be taken care of now that you've changed the StatementIndexer. It just depends whether you added a migration for the statement ref timestamps and whether the refs are created with the correct timestamps (which I think they should be because it just copies the referred statement).

PR #783, opened 2016-02-23
Forces composer to use the forked version of jenseggers/mongodb for much improved counts

ht2:


ht2:
Going to try this on a fresh PR

PR #816, opened 2016-05-18
Better hinting for counts

ht2:


app/controllers/LrsController.php
```diff
@@ -190,14 +190,15 @@ public function destroy($lrs_id){
    */
   public function statements($lrs_id){
     $site = \Site::first();
+    return "hello";
```
ryansmith94: @ht2 was this an accident?

ht2: YES!

PR #817, opened 2016-05-18
add Event::fire('Statements.store') in createStatements

fsieduc:
Hi,

We are planning a LearningLocker - RabbitMQ integration.

We need a hook in StatementStoreController::createStatements().

May be there are more places to hook...

Thanx for this great software.

### DIFF
--- a/app/controllers/xapi/StatementStoreController.php
+++ b/app/controllers/xapi/StatementStoreController.php
@@ -131,13 +131,18 @@ class StatementStoreController {
     }
 
     // Saves $statements with attachments.
-    return $this->statements->store(
+    $rc = $this->statements->store(
       $statements,
       is_array($parts['attachments']) ? $parts['attachments'] : [],
       array_merge([
         'authority' => $this->getAuthority($options['client'])
       ], $options)
     );
+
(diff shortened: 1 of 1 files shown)


ht2:
Hey @fsieduc,

Thanks for this PR - I don't think there is any need for us to commit this into core at this point in time, as all it will do is pass around an extra event (with a potential hefty amount of data). I see you already have this committed to your fork, so probably best you keep installing and cloning from your forked repo.

Good luck with the RabbitMQ integration though! We've had some good success utilising Redis to publish events about statement insertions, so I can see why you'd be finding this useful.

fsieduc:
Thanks for your quick reply :-)

Ok, we will continue to use our fork until you eventually merge this new event.

> with a potential hefty amount of data

You're right!
I think this point can be optimized by passing a reference of the $statements (&$statements) instead of copying it.

    \Event::fire('Statements.store', array(&$statements));

I've tested it and works like a charm.

Do you want another PR ?

ht2:
Good thinking re:references. However, I think until another part of the app is utilising the event we'd have no need to bring it in as core. Would love to see contributions to the app that would utilise such a feature however!

Issue #754, opened 2015-10-23
CORS OPTIONS request not working


chadabaker:
**Version**
1.9.0

**Steps to Reproduce**
1. Send OPTIONS request from HTTPS application domain to HTTPS learning locker domain.
2. See that request fails due to missing header (Access-Control-Allow-Origin) 

**Expected behaviour**
Response should include "Access-Control-Allow-Origin" header

**Actual behaviour**
Response does not include header

**Server information**
CentOS, MongoDB, Basic Auth

**Client information**
OS: Windows and Mac
Browser: All
Content: Adobe Captivate 8, published for Tin Can
Content (cont): Seems to be sending requests through scormdriver.js

**Additional information**
Route configuration includes the following:
Route::options($route->getUri(), 'Controllers\API\Base@CORSOptions');

I'm not sure where the CORSOptions function is declared. Seems that it was removed in commit e1ac0d5a76a3ec415f9d12abbccb3c0ee4fe525c and not included in commit 666a7c6f6baffb399f201e796b0c19f73374d2a3. 

**Full Error**
XMLHttpRequest cannot load https://learninglocker.mydomain.com/data/xAPI/activities/state?stateId=suspen…%3A%22Chad%20Baker%22%7D&registration=7c624b02-84d0-47dc-a4a3-0ea26c5a1651. Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'https://otp.mydomain.com' is therefore not allowed access.

ht2:
This header is set in the https://github.com/LearningLocker/learninglocker/blob/master/app/filters.php#L25 file. It would seem to me that perhaps there is no HTTP_ORIGIN being sent to allow it to create the `Access-Control-Allow-Origin` header perhaps? Can you see if it is present in the request headers? Any cross origin request should be sending this.

chadabaker:
Here are the request headers:

OPTIONS /data/xAPI/activities/state?stateId=bookmark&activityId=http%3A%2F%2FCourse_ID1&agent=%7B%22objectType%22%3A%22Agent%22%2C%22account%22%3A%7B%22name%22%3A%22shdrlevel90en%22%2C%22homePage%22%3A%22https%3A%2F%2Fotp-shdr.mydomain.com%22%7D%2C%22name%22%3A%22First%20Last%22%7D&registration=3a538fc8-c06c-46f7-b2d7-3e3e129e8427 HTTP/1.1
[...]

PR #848, opened 2016-08-01
Remove custom CORS routes

ht2:
The barryvdh/laravel-cors package should be handling all pre-flight requests.


ryansmith94:
#754

Issue #853, opened 2016-08-17
I have problem with the "create report" (reports section) and "add export" (export section) buttons.


dimis283:
You need to replace the parts of this template where you see {{some text}}.

**Version**
{{branch}} at {{commit}} on {{version - found in your copy of the VERSION file}} 

**Steps to reproduce the bug**
1. {{step 1}}
2. {{step 2}}

**Expected behaviour**
{{feature}} should be {{expectedResult}} because {{reason}}.

**Actual behaviour**
{{feature}} is {{actualResult}}.

**Additional information**
OS: {{operatingSystem}}
Browser: {{browser}} version 1.0.1
{{additionalInfo}}

dimis283:
When I push them nothing happens.3 days before the  "create report" button was working. Now I have the error (using the "create report" button) at console of chrome "TypeError: Cannot read property 'id' of undefined" What is wrong?

ryansmith94:
Hi @dimis283, please update your initial comment.

> You need to replace the parts of this template where you see {{some text}}.

hallshouse:
I can confirm this issue.
After upgrading to v1.3.4, the Create Report button in the reporting view does not function.
After upgrading to v.1.3.4 the Add export button in the exporting view does append the URL with "#new", but does nothing.

ryansmith94:
Sorry @hallshouse, did you mean v1.13.4?

hallshouse:
Thanks for correcting me, I had actually entered v1.3.4 (my mistake).

now I've corrected that by redoing the upgrade procedure found at: http://docs.learninglocker.net/upgrading/ and using v1.13.4

I've run into an issue in doing so:
I'm now getting the following error in the statements view:
Class '\Locker\XApi\Str' not found
It's pointing to: 43.          Helpers::validateAtom(new $class($value));

Any idea what this could be?

hallshouse:
Disregard, I failed to run the:
git rm --cached app/config/local -r; git add app/config/local; git commit -m 'Fixes local config.'; php -r "readfile('https://getcomposer.org/installer');" | php; php composer.phar install
Closed.
Getting more coffee!

ryansmith94:
No worries @hallshouse 👍 

Ha awesome! Thanks, I was literally about to start replying to that issue. Enjoy your ☕

hallshouse:
@ryansmith94 To clarify, my stupidity on the upgrade is closed, but the Create Report and Add export buttons issue is still open.

[...]

PR #864, opened 2016-09-09
Fixes #806.

### LINKED ISSUE #806: Export page should provide info if no reports are available
The Export page allows you to add new Exports even if a report doesn't exist in the LRS. 

Some form of explanation (and link?) to go and make a report first should be provided and the "Add Export" button should not be available to the user.

As noted in #805

ryansmith94:
#806 
Now displaying an error message when you attempt to create a export without a report.


ryansmith94:
May also resolve #853

Issue #910, opened 2017-03-23
Add ability to set name, mbox, and scopes when creating client through API


ryanrolds:
**Motive**
We need to be able to create clients with specific names, authority, and scopes through the API.

**Result**
The purposed changes: https://github.com/irisEd/learninglocker/commit/3a592cf0b535659d95514b63706758e32c44c048

PR #911, opened 2017-03-23
Add ability to set name, mbox, and scopes when creating client through API

ryanrolds:
https://github.com/LearningLocker/learninglocker/issues/910

### DIFF
--- a/app/controllers/xapi/BasicRequestController.php
+++ b/app/controllers/xapi/BasicRequestController.php
@@ -40,11 +40,11 @@ class BasicRequestController extends BaseController {
     if (!is_array($scopes)) {
       return \Response::json([
         'error' => true,
-	'success' => false,
+        'success' => false,
         'message' => 'Scopes must be an array or not defined',
-	'code' => 400
+        'code' => 400
       ], 400);
-    } 
+    }
 
     $client = $this->client_repo->store([
       'authority' => [
--- a/app/controllers/xapi/BasicRequestController.php
+++ b/app/controllers/xapi/BasicRequestController.php
@@ -10,6 +10,10 @@ class BasicRequestController extends BaseController {
   // Defines properties to be set by filters.
   protected $params, $lrs;
 
+  // Sets constants for param keys.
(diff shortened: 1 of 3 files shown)


app/controllers/xapi/BasicRequestController.php
```diff
@@ -19,28 +23,43 @@ class BasicRequestController extends BaseController {
...
   public function store(){
-  	$opts = ['lrs_id' => $this->lrs->_id];
-  	$client = $this->client_repo->store([
+    $opts = ['lrs_id' => $this->lrs->_id];
+    $name = \LockerRequest::getParam(self::CLIENT_NAME, 'API Client');
+    $mbox = \LockerRequest::getParam(self::CLIENT_MBOX, 'mailto:hello@learninglocker.net');
+    $scopes = \LockerRequest::getParam(self::CLIENT_SCOPES, ['all']);
+
+    // Scopes must be an array or they get a 400
+    if (!is_array($scopes)) {
```
ryansmith94: Nice

app/locker/repository/Client/EloquentRepository.php
```diff
@@ -63,7 +64,7 @@ protected function constructStore(Model $model, array $data, array $opts) {
     ];
     $model->lrs_id = $opts['lrs_id'];
     $model->authority = $data['authority'];
-    $model->scopes = ['all'];
+    $model->scopes = $data['scopes'];
```
ryansmith94: Ah this needs to default to ['all'] or the [ClientController@create](https://github.com/LearningLocker/learninglocker/blob/develop/app/controllers/ClientController.php#L77) needs to change to supply the scopes. The ClientController provides another API for creating clients although I believe it can't be used outside of the UI because of the auth required.

ryanrolds: Does L57 and L16 not handle this correctly?

ryansmith94: Ah yeah sorry, good spot

PR #943, opened 2017-09-19
Only allow valid selection of roles.

cbishopvelti:


lib/constants/orgScopes.js
```diff
@@ -1,4 +1,11 @@
 import { ALL } from 'lib/constants/scopes';
+import {
```
andrewhickey: Lodash not used in this file?

andrewhickey: Looks like this needs rebasing from develop for the build to pass. Also probably failing linting. Otherwise looks good :)

PR #953, opened 2017-09-25
Issue ll 343 Logs out a user if a request returns unauthorised.

cbishopvelti:


ui/src/redux/modules/aggregation/fetchAggregation.js
```diff
@@ -32,7 +32,7 @@ const aggregationShouldFetchSelector = pipeline => createSelector(
 
 const fetchAggregaation = createAsyncDuck({
   actionName: 'learninglocker/aggregation/FETCH_AGGREGATION',
-  failureDelay: 20000,
```
andrewhickey: What's the thoughts behind lowering this?

PR #979, opened 2017-10-10
Prevents repeat failed requests

cbishopvelti:


cbishopvelti:
Breaks deleting visualisations.

andrewhickey:
For reference, this PR is not breaking, the deletion problem was pre-existing and has been solved.

PR #992, opened 2017-10-13
Fixed non super admin users unable to create new organisation.

cbishopvelti:
- Fixes shareable roles on statement forwarding
- Fixed non super admin users unable to create new organisation.
- Fixes roles for different token types


ht2:
Need to make same changes we did for uploadLogo scoping, to the downloadLogo changes

ht2:
Hooray.

Still changes needed to how clients vs org tokens respond on some models.

PR #997, opened 2017-10-18
Filtered selects on CRUD GET

ht2:


lib/models/plugins/addCRUDFunctions.js
```diff
@@ -200,7 +204,34 @@ async function getConnection({
...
+    } else {
+      finalProjection = project;
+    }
+  } else {
+    finalProjection = scopeSelect;
+  }
+
+  if (finalProjection) {
+    query.select(finalProjection);
+  }
```
andrewhickey: Looks good as long as getScopeSelect returns falsy by default

lib/services/auth/selects/getScopeSelect.js
```diff
@@ -1,9 +1,9 @@
...
 
-export default async (opts) => {
-  switch (opts.modelName.toLowerCase()) {
+export default async ({ modelName, actionName, authInfo }) => {
+  switch (modelName.toLowerCase()) {
     case 'user':
-      return getUserScopeSelect(opts);
+      return getUserScopeSelect({ actionName, authInfo });
     default:
       // return the default "null" which will allows all selects
```
andrewhickey: It does :)

PR #1027, opened 2017-11-09
Feature personas query b

cbishopvelti:


lib/helpers/convert$personaIdent.js
```diff
@@ -8,36 +10,46 @@ import { getConnection } from 'lib/connections/mongoose';
...
+      maxTimeMS: MAX_TIME_MS,
+      maxScan: MAX_SCAN
+    });
+    attributes = map(result.edges, ({ node }) => node);
+  } catch (err) {
+    logger.error('convert$personaIdent', err);
+    return;
+  }
+
+  return { $in: map(attributes, ({ personaId }) => personaId) };
```
ht2: We had issues here where a `null` value (wrongly placed) onto idents previously would match on all statements that also did not have a persona. Could we do some similar filtering to before to remove undefined/null values to ensure we've got a clean set of values passed out to our `$in` query

PR #1029, opened 2017-11-10
Enable .env config to override system email address

eashman:
Most email gateways will not allow using a non-domain address for the "from" in outbound emails they deliver.  This prevents the emailer working in Learning Locker.

This config change will let you override the default name/address in the config to enable delivery.

### DIFF
--- a/lib/helpers/email.js
+++ b/lib/helpers/email.js
@@ -1,11 +1,13 @@
 import logger from 'lib/logger';
 import transporter from 'lib/connections/nodemailer';
 import _ from 'lodash';
+import defaultTo from 'lodash/defaultTo';
 import fs from 'fs';
 import htmlToText from 'html-to-text';
 
 const defaultMailOptions = {
-  from: `${process.env.LL_FROM_NAME} <${process.env.LL_FROM_EMAIL}>` || 'Learning Locker <noreply@learninglocker.net>'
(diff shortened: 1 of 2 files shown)


ryansmith94:
Hey @eashman, thanks for the PR, code changes look fine, I'll leave this to @ht2 to merge 👍

lib/helpers/email.js
```diff
@@ -5,7 +5,7 @@ import fs from 'fs';
 import htmlToText from 'html-to-text';
 
 const defaultMailOptions = {
-  from: 'Learning Locker <noreply@learninglocker.net>',
+  from: `${process.env.LL_FROM_NAME} <${process.env.LL_FROM_EMAIL}>` || 'Learning Locker <noreply@learninglocker.net>'
```
ht2: Not sure this is going to work as expected, as empty values on both would lead to an string of `" "` which would not be falsey. Perhaps look at `lodash/defaultTo` instead?

eashman: The problem doing it similarly to some of the other variables is you don't want to introduce another required default in the config file.   

Is there an example of how you've used defaultTo elsewhere in the code?

ht2:
I'll need to update the docs accordingly
https://github.com/LearningLocker/docs/blob/master/src/guides-configuring.md

PR #1030, opened 2017-11-13
Map &46; to . On retrieval.

cbishopvelti:


ht2:
Just want to check that this method is also covered: https://github.com/LearningLocker/learninglocker/blob/develop/lib/models/statement.js#L150

PR #1040, opened 2017-11-21
LL-426 Migrates statements to contain properties for refs

ryansmith94:
This allows us to merge LearningLocker/xapi-service#34 which includes performance improvements for statements. Before this is merged, the deploy script needs changing to run the migrations.


ryansmith94:
In terms of the migration command. I think it would be wise to simplify things a little bit.

I think we could split this into two commands (migrate and rollback) to match what other migration tools have. The migrate command could take a "migration key" to run a single migration, or migrate all outstanding migrations if a key isn't given. The rollback command could take a "migration key" to run a single migration and run the last migration if a key isn't given.

I think we should also remove the upFn from the database, realistically it won't actually enforce anything since the up function will usually call other functions. Just relying on the "migration key" would be consistent with other migration tools.

ht2:
Merged into #1041

PR #1048, opened 2017-11-27
LL-441 Personas migration

ryansmith94:
This isn't tested yet @ht2, feel free to change the base.

Essentially, there are two migrations. One to remove identifiers from personas because they're redundant, and a second to create attributes from personaIdentifiers and update the fields on personaIdentifiers.

The second migration, streams through the identifiers first creating attributes in batches via bulk ops. After that, it removes unused fields (scores and idents) and renames the `uniqueIdentifier` field to `ifi`.


ht2:
Closing and remaking this PR somewhere else as this change list is borked!

PR #1058, opened 2017-12-08
API amends

ht2:


api/src/controllers/utils/handleError.js
```diff
@@ -10,28 +10,55 @@ import logger from 'lib/logger';
 import EmptyCsvError from 'lib/errors/EmptyCsvError';
 import DuplicateCsvHeadersError from 'lib/errors/DuplicateCsvHeadersError';
 
+import PersonaConflict from 'personas/dist/errors/Conflict';
+import NoModel from 'jscommons/dist/errors/NoModel';
+import PersonaNoModelWithId from 'personas/dist/errors/NoModelWithId';
+
 export default (res, err) => {
   logger.error(err);
   if (isNull(err) || isUndefined(null)) return res.status(500).send('Error');
```
ryansmith94: This bit could potentially be moved to the bottom of this function and we could send 'Error' as the message every time just in case the actual error caught exposes some internal things.

PR #1070, opened 2017-12-19
v2.1.1

ht2:


lib/models/statement.js
```diff
@@ -6,7 +6,8 @@ import defaultTo from 'lodash/defaultTo';
 import boolean from 'boolean';
 import Promise from 'bluebird';
 import keys from 'lodash/keys';
-import isObject from 'lodash/isObject';
+import isPlainObject from 'lodash/isObject';
```
ryansmith94: lodash/isObject should be lodash/isPlainObject

PR #1083, opened 2018-01-08
LL-433 Google Cloud file storage

ht2:


lib/services/files/storage/index.js
```diff
@@ -1,12 +1,14 @@
 import * as rackspace from './rackspace';
 import * as amazon from './amazon';
+import * as google from './google';
 import * as local from './local';
 
 let fsRepo;
 
 switch (process.env.FS_REPO) {
```
ryansmith94: I'm not massively sure why this isn't a function that just returns the fsRepo, then we do `const fsRepo = getFSRepo();`

PR #1086, opened 2018-01-09
LL-452 Google PubSub for worker 

ht2:


.env.example
```diff
@@ -141,6 +141,10 @@ QUEUE_NAMESPACE=DEV
 #AWS_SQS_SECRET_ACCESS_KEY=
 #AWS_SQS_DEFAULT_REGION=
 
+# GOOGLE CLOUD PUBSUB creentials
+#GOOGLECLOUD_PUBSUB_KEY_FILENAME=
```
ryansmith94: Can we make this consistent either `GOOGLE_CLOUD` or `GOOGLECLOUD`?

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ Format based on [Keep a Changelog](http://keepachangelog.com/)
   - Google cloud services 
     - Storage
     - PubSub queues
+      - Requires PubSub API to be enabled for your Google Cloud project
+      - New environment configs:
+        - `GOOGLECLOUD_PUBSUB_KEY_FILENAME` 
+          - Path to your Service account key JSON file. Must be configured to allow read/write/delete to your Cloud Storage bucket
+        - `GOOGLE_CLOUD_PUBSUB_PROJECT_ID` - Project ID
```
ryansmith94: Same consistency issues here.

ryansmith94:
Not a massive fan of the amount of callbacks being used here instead of promises, but I appreciate that it would be a large refactor.

.env.example
ht2: Yea that's just wrong - my bad. Should have been GOOGLE_CLOUD (and is consistent with the storage as well)

PR #1089, opened 2018-01-12
LL-463  [Personas] Change import state when processing

cbishopvelti:


ht2:
Looks like import-persona tests are broken

PR #1090, opened 2018-01-15
LL-468 Rewrite "statement.timestamp" in aggegation requests -> v2.1.3

ht2:


ht2:
https://ht2labs.myjetbrains.com/youtrack/issue/LL-468

lib/helpers/convertStatementTimestamp.js
```diff
@@ -0,0 +1,92 @@
...
+  isString,
+  identity,
+  mapValues,
+} from 'lodash';
+
+const createFilterParser = (key, parser) => (filter) => {
+  let newFilter;
+  if (has(filter, key)) {
+    try {
+      newFilter = parser(filter);
```
ht2: Had to change this to not return the filter straight away, as we still needed it to go and map through the object

ryansmith94: This should probably be part of the isPlainObject block.

PR #1096, opened 2018-01-19
Issue ll 455b (multiple Shareable Dashboards Links)

cbishopvelti:


ht2:
Need to work on filtering journeys by `person._id` if included in filter

PR #1102, opened 2018-01-30
Feature personas workers

cbishopvelti:


lib/models/importPersonasLock.js
```diff
@@ -0,0 +1,13 @@
+import mongoose from 'mongoose';
```
ryansmith94: Let's check this model is needed.

cbishopvelti: It's needed

ryansmith94: Feels like this is something that the persona service locks should handle?

Issue #1107, opened 2018-02-12
Bug: Sending email via secure SMTP is never possible


davidpesce:
**Version**
2.1.3 

**Steps to reproduce the bug**
1. Configure SMTP settings in .env file
2. Restart pm2 services with --update-dev
3. Attempt to send email
4. Check api-stderr log and you'll see that it's trying to send it as anonymous.

**Expected behaviour**
Secure email setting in .env should be recognized.

**Actual behaviour**
Email sending fails because of non-secure

**Additional information**
OS: CentOS 7.4 (64)
Browser: Chrome (latest)

I believe the issue is here: https://github.com/LearningLocker/learninglocker/blob/16bc883bb4af301d74d503d34677ca4c04aa6def/lib/connections/nodemailer.js#L7

davidpesce:
Confirming this is fixed in 2.1.4. Closing issue.

PR #1108, opened 2018-02-12
fix for sending secured smtp emails

davidpesce:
Small fix for sending secured smtp emails. Current .env was being ignored. Fixes issue #1107


ryansmith94:
Hey thanks @davidpesce, just restarted that Travis build, think it was just a weird Travis error.

PR #1127, opened 2018-03-26
LL-63 Fixed 401 not logging out

cbishopvelti:


ht2:
Issue with the jasmine test here @cbishopvelti?

PR #1129, opened 2018-04-06
fixed smtp options to skip auth if auth params are not defined

davidscholberg:
Heyyyyyy,

This pull request fixes an issue I encountered where SMTP authentication cannot be turned off via configuration. We have an internal SMTP server that we allow unauthenticated access to, so I wanted to be able to use it without needing any credentials.

Nodemailer will always try to use SMTP authentication if the `auth` option value is defined ([as documented here](http://nodemailer.com/smtp/#authentication)), even if the `user` and `pass` values are not defined. So the basic idea of this PR is that the `auth` object is only added to the `smtpConfig` object if either `SMTP_USER` or `SMTP_PASS` are defined in the `.env` file.

Let me know if there's anything else I can do. Thanks.


davidscholberg:
Is there anything I can do to help move this along?

Ian247:
@davidscholberg I will email you directly about this.

caperneoignis:
Oh I also did this because I didn't see a issue in github. #1307 is the pull request I did, and everything is passing.

ht2:
Closing as merged in via #1307

PR #1131, opened 2018-04-09
LL-562 Upgraded to node v8.11.1

cbishopvelti:
Requires global upgrade of pm2


ht2:
@asahd Strange request, but could you try QAing this branch running Node 6?

PR #1134, opened 2018-04-10
Statement forwarding now decodes &46; to .

cbishopvelti:


ht2:
https://github.com/LearningLocker/learninglocker/blob/master/lib/models/statement.js#L236

Might need exporting?

PR #1138, opened 2018-04-11
LL-445 Ensure statementForwarding query is valid json.

cbishopvelti:


ht2:
Check that UI is sending valid JSON

PR #1141, opened 2018-04-13
LL-522 Use local file when importing csvs, and use clamd

ht2:


ht2:
Requires this to be merged once released: https://github.com/LearningLocker/deploy/pull/51

PR #1177, opened 2018-06-07
LL-551 Shareable dashboards filter parameter.

cbishopvelti:


ht2:
Missing functionality: 

An additional setting of "Filter required: Yes/No" should be available when "Any" or "Secured" are selected. This will prevent the dashboard returning any data, unless a valid filter is present in the filter query param.

ht2:
https://github.com/LearningLocker/learninglocker/pull/1194

PR #1181, opened 2018-06-11
LL-187 Adding an actor in the object slot of the query builder causes things to break

cbishopvelti:


ryansmith94:
@asahd any idea what held this one up?

PR #1189, opened 2018-06-19
[DO NOT MERGE] refactor: Visualisations refactor.

ryansmith94:
There is some weirdness in the sense that we convert the models to a new format client side, the plan is to eventually migrate the models in the database and server side, but right now we'd like to avoid the breaking change in the API.

Goals
- Want to keep all of the components and UI logic for visualisations in one folder this time (this will allow extraction later if we decide to go have an LL Insights thing).
- Want to keep each visualisation "self-sufficient" in the sense that it makes API calls to LL via the `withVisualisationResults` HOC so that we can extract the components into other projects (i.e. LL Insights) later.
- Want to allow switching between visualisations within reason (can't switch from a "values" visualisation like a bar chart to a "tuples" visualisation like a scatter chart).
- Want to make the charts simpler so that we can easily add new things to the charts like average lines, median lines, etc.
- Want to make it easier to add new visualisations later.

Definitions
- Counter Visualisation = Visualisation that expects a result array with one object with a single count.
- Values Visualisation = Visualisation that expects a result array with zero or more objects containing a count, model, and group.
- Tuples Visualisation = Visualisation that expects two result arrays with zero or more objects containing a count, model, and group. The two result arrays are combined by grouping with the `group` in the result objects to get a single result array with two counts (hence using the word "tuple").

[Original Google Doc](https://docs.google.com/document/d/1U0ptO4-rrx4Gju64voCS_hHAPHOmx2dVBTyvWLalR4k/edit)


ht2:
Superceded by https://github.com/LearningLocker/learninglocker/pull/1207

PR #1205, opened 2018-07-16
Visualization soft update pt2

ht2:


Ian247:
piechartnan
 Pie Chart showing NaN and large segment is missing.

Ian247:
journeyerror
Cannot choose a journey visualisation, this error throws when try and gets stuck trying to save selection. On refresh still unselected.

PR #1213, opened 2018-07-30
LL-595 Refactor query to use $in

cbishopvelti:
Closes https://github.com/LearningLocker/enterprise/issues/888
https://ht2labs.myjetbrains.com/youtrack/issue/LL-595


cbishopvelti:
@asahd This is an existing issue

ryansmith94:
Are we still using the top level `$and` with the `$in`/`$nin`? If we aren't then I think we probably need to because of actor accounts, really we should find a way to query the actors by a single field. Not sure what the performance implications are of mixing `$or` into these queries for the actors or even wrapping the `$and`, probably not good to be honest.

Definitely need some database migrations to migrate existing queries to use `$in`/`$nin` to move this forward. You could potentially wrap `$in`/`$nin` inside the existing `$or` to avoid the need to remove migrations for now, but the queries wouldn't get the performance improvement so I don't think we would want to do that for the purpose of this ticket.

Guessing this has lost priority for now hence the stale branch.

h-kanazawa:
I fixed LL-595 on #1338

PR #1218, opened 2018-08-01
v2.4.2

ht2:
* fix(pm2): Garbage collection for worker.json

* fix(pm2): GC for worker.json.dist

* fix(pm2): GC for all.json

* fix(pm2): GC for all.json.dist

* refactor(worker): Drop completed job log to debug


ht2:
Need to remove max memory on workers now that we have garbage collection

PR #1276, opened 2018-11-12
v2.6.3

ht2:
### Fixed
  - Counter scale and centering
  - Allow merging from non-existent persona
  - Allow upsert for toPersona on import
  - Update LRS statement count one by one


ht2:
@asahd Don't really need you to review this one I guess - its just got some persona import fixes (when scale gets crazy: 200k+)

PR #1292, opened 2019-01-07
Keep $sample as first in aggregate pipeline, for efficiency fixes 1291

davidavdav:
This PR makes sure that in an aggregate query, a `$sample: { size: size }` statement that occurs as a first in the pipeline stays the first in the pipeline.  This is important for the way mongo executes the sample, when it is a first, the collection is really sampled, [otherwise a full scan is made](https://docs.mongodb.com/manual/reference/operator/aggregation/sample/#pipe._S_sample), and the result is sorted randomly, which is way less efficient.

### DIFF
--- a/lib/models/statement.js
+++ b/lib/models/statement.js
@@ -179,6 +179,11 @@ schema.statics.aggregateByAuth = function aggregateByAuth(
     const scopedFilter = await getScopeFilter({ modelName, actionName, authInfo });
     const organisation = getOrgFromAuthInfo(authInfo);
 
+    let sampleStatement = null;
+    if (parsedPipeline.length > 0 && parsedPipeline[0]['$sample']) {
+      sampleStatement = parsedPipeline.shift();
+    }
+
     parsedPipeline.unshift({
       $match: scopedFilter
     });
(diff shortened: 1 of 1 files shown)


asahd:
@ryansmith94 Probably something that makes more sense for @ht2 to test? Since this is nicely isolated from anything else and I don't think we use `$sample` in any aggregations from the UI, etc.

ht2:
We do not use $sample, and I can see why this would be awkward. It is worth noting that in a truly multi tenant system, it would be possible to run an aggregation with $sample that may return no records (or a non-representative sample) due to the organisation _always_ being introduced via a $match as a (top level) pipeline stage

I'm actually tempted to suggest that the addition of a $sample stage should be done via a query param to the aggegation endpoint, and then automatically added as the front stage by the system when asked for.  This would enable a developer to actually use the $sample stage further down the pipeline, which may be the intended use case (albeit a slow one)

ryansmith94:
Yeah I like the idea of the sample query parameter. I think we should go with that.

ht2:
Feel free to resubmit with the sample added as a query param option - for the moment, closing this

davidavdav:
Sorry, I did not understand.  Am I supposed to reimplement this as a query parameter?

ht2:
Yes!

ht2:
I don't want to force a change to the pipeline the client has submitted, rather than add an optional query param (boolean) that would insert a $sample stage at the front if `sample=true` was passed. It could apply a sensible default (set in the .env) but could also be overwritten via another query param such as `sampleSize=10000`

PR #1297, opened 2019-01-22
LL-345 Fix grouping by Persona.Import in visualisations

### LINKED ISSUE #1204: Can't group by persona attributes in visualisations
Not sure if user error or bug, but:

**Version**
master at 6b58feb  on 2.4.1

**Steps to reproduce the bug**
1. Import CSV or manually add attributes to Persona
2. Go to Visualisation
3. Select any graph type, for an example Bar
4. in Y axis use any imported field, e.g. LL suggests persona.import.FieldName

**Expected behaviour**
Expect LL to build a graph based on the imported field, for an example Country of the personas. Show number of statements per persona's country. Country field and data successfully imported and are visible in personas attributes and are successfuly used as query filters. 

**Actual behaviour**
Visualisation outputs No data in this time range.

**Additional information**
OS: Windows 10
Browser: Firefox 61.0.1 (64-bit
All other visualisations and etc work as expected.

h-kanazawa:
Fix #1204 
https://ht2labs.myjetbrains.com/youtrack/issue/LL-345

- [ ] performance tests


ui/src/utils/visualisations/aggregationGroup/grouper/index.js
```diff
@@ -78,38 +78,69 @@ const getExistsMatch = ({ valueType, groupType, valueOpCase }) => {
...
+  const unwindStage = createStagePipeline('$unwind', {
+    path: '$personaAttrs'
+  });
+
+  const matchAttrsKeyStage = createStagePipeline('$match', {
+    'personaAttrs.key': groupType.split('.')[2] || ''
+  });
+
+  return lookupStage
+    .concat(unwindStage)
```
ht2: I think an additional match stage before the unwind will actually filter any people's records that don't have the attribute prior to unwinding. Debatable if this make the following two steps quicker or not

```suggestion
    .concat(matchAttrsKeyStage)
    .concat(unwindStage)
```

ui/src/utils/visualisations/aggregationGroup/grouper/index.spec.js
```diff
@@ -1,13 +1,79 @@
...
+      valueType: 'statements',
+      groupType: 'persona.import.abcde',
+      operatorType: 'uniqueCount'
+    });
+
+    test('pipeline has correct stages', () => {
+      expect(result.toJS().map(stage => Object.keys(stage)[0]))
+        .toEqual([
+          '$lookup',
+          '$unwind',
```
ht2: Required if implementing my above suggestion

```suggestion
          '$match',
          '$unwind',
```

ui/src/utils/visualisations/aggregationGroup/grouper/index.js
h-kanazawa: OK. I'll measure performance with the fake data.

ryansmith94:
Note to self:
- [ ] Update docs to state new minimum version of Mongo is 3.6.
- [ ] Ensure install script uses Mongo 3.6.
- [ ] Add note to release notes to say this requires 3.6.

PR #1300, opened 2019-01-25
LL-703 feat: Adds option to forward attachments.

ryansmith94:
https://ht2labs.myjetbrains.com/youtrack/issue/LL-703


ryansmith94:
Ideally this could use a new test which would require uploading an actual file via the xapi-statements repo and checking that the correct stream was returned.

ryansmith94:
Added a test but been unable to get the mocking library to correctly match on the headers. I have logged the headers from the request out and seen that it is uses the correct headers and the correct body in the request.

ryansmith94:
Will spin this up tomorrow and do a little QA on it via the UI to make sure everything is working as expected, if I remember I'll add some screenshots in here too.

ryansmith94:
screenshot 2019-02-19 at 17 01 10

ryansmith94:
Got this setup with RequestBin to test that it sends statements on their own as expected.
screenshot 2019-02-19 at 17 21 38

ryansmith94:
Tested with RequestBin and an attachment too.
screenshot 2019-02-19 at 17 40 31

worker/src/handlers/statement/statementForwarding/getStatementsRepo.js
```diff
@@ -0,0 +1,61 @@
...
+      },
+    },
+    events: {
+      facade: process.env.EVENTS_REPO,
+      redis: {
+        client: createClient(),
+        prefix: process.env.REDIS_PREFIX,
+      },
+      sentinel: {
+        client: createClient(),
```
ht2: Will this create a Sentinel client even if its not configured to need it?

ryansmith94: Because of how that function works it would have just passed in either a Redis or Sentinel client depending on what was needed and not used it. I've updated this code to just pass null connections through since we don't actually need them right now anyway.

PR #1301, opened 2019-01-29
ci(cypress): Attempts to remove video recording to speed up build.

ryansmith94:
Seems to reduce build time by 1 minute. Please review and merge @ht2.


ryansmith94:
Thanks for reviewing @ht2, can merge when ready.

PR #1307, opened 2019-01-30
updated to make sending auth optional

caperneoignis:
Updated to resolve issue 1306. And added additional keys to the .env.example file to allow people to find the additional settings later. This is to prevent nodemailer from sending a plain AUTH request to a relay or smtp server that does not have any auth types listed, IE "205-AUTH" instead of "205-AUTH PLAIN LOGIN". For some reason nodemailer does not automatically do this if username and password is set to blank, I assume they have their reasons.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -12,13 +12,13 @@ SITE_URL=http://127.0.0.1
 UI_PORT=3000
 
 # If UI is on a different host
-#UI_HOST
+#UI_HOST=127.0.0.1
 
 # Expose the API on this port
 API_PORT=8080
 
 # If API is on a different host.
-#API_HOST
+#API_HOST=127.0.0.1
 
 # Use an additional port for unit tests
 #TEST_API_PORT=808080
@@ -135,9 +135,9 @@ GOOGLE_ENABLED=false
 #SMTP_REQUIRE_TLS=false
 
 # If your SMTP server/relay does not have SMTP AUTH,
-# IE treats all local connections as already authenticated.
-# Then set this to false, default is true you do want SMTP AUTH.
-#SMTP_AUTH_REQ=true
+# (i.e. treats all local connections as already authenticated),
+# then set this to false, default is true you do want SMTP AUTH.
+#SMTP_AUTH_REQUEST=true
 
 #LL_FROM_NAME="Learning Locker"
 #LL_FROM_EMAIL=noreply@learninglocker.net
--- a/.env.example
+++ b/.env.example
@@ -11,8 +11,14 @@ SITE_URL=http://127.0.0.1
 # Expose the UI on this port
 UI_PORT=3000
 
+# If UI is on a different host
+#UI_HOST
+
 # Expose the API on this port
 API_PORT=8080
+
+# If API is on a different host.
+#API_HOST
 
 # Use an additional port for unit tests
 #TEST_API_PORT=808080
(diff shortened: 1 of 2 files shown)


.env.example
```diff
@@ -128,6 +134,11 @@ GOOGLE_ENABLED=false
 #SMTP_IGNORE_TLS=false
 #SMTP_REQUIRE_TLS=false
 
+# If your SMTP server/relay does not have SMTP AUTH,
+# IE treats all local connections as already authenticated.
+# Then set this to false, default is true you do want SMTP AUTH.
+#SMTP_AUTH_REQ=true
```
asahd: This name is confusing, considering that it's right under `SMTP_REQUIRE_TLS` there is a bias to read this as `SMTP_AUTH_REQUIRE`, which isn't the intended meaning.

Should be renamed for clarity

```suggestion
#SMTP_AUTH_REQUEST=true
```

.env.example
```diff
@@ -11,9 +11,15 @@ SITE_URL=http://127.0.0.1
 # Expose the UI on this port
 UI_PORT=3000
 
+# If UI is on a different host
+#UI_HOST
```
asahd: Variables should have example values
```suggestion
#UI_HOST=127.0.0.1
```

.env.example
```diff
@@ -11,9 +11,15 @@ SITE_URL=http://127.0.0.1
...
+
 # Expose the API on this port
 API_PORT=8080
 
+# If API is on a different host.
+#API_HOST
```
asahd: Variables should have example values
```suggestion
#API_HOST=127.0.0.1
```

.env.example
```diff
@@ -128,6 +134,11 @@ GOOGLE_ENABLED=false
 #SMTP_IGNORE_TLS=false
 #SMTP_REQUIRE_TLS=false
 
+# If your SMTP server/relay does not have SMTP AUTH,
+# IE treats all local connections as already authenticated.
```
asahd:
```suggestion
# (i.e. treats all local connections as already authenticated),
```

.env.example
```diff
@@ -128,6 +134,11 @@ GOOGLE_ENABLED=false
...
+# Then set this to false, default is true you do want SMTP AUTH.
```
asahd:
```suggestion
# then set this to false, default is true you do want SMTP AUTH.
```

lib/connections/nodemailer.js
```diff
@@ -8,21 +8,32 @@ const smtpConfig = {
...
-  requireTLS: boolean(defaultTo(process.env.SMTP_REQUIRE_TLS, false)),
-  auth: {
-    user: process.env.SMTP_USER,
-    pass: process.env.SMTP_PASS
-  }
+  requireTLS: boolean(defaultTo(process.env.SMTP_REQUIRE_TLS, false))
 };
 
+// Placed like this since the boolean set default can be hard to read in an if condition.
+const sendAuthRequest = boolean(defaultTo(process.env.SMTP_AUTH_REQ, true));
```
asahd: As above

```suggestion
const sendAuthRequest = boolean(defaultTo(process.env.SMTP_AUTH_REQUEST, true));
```

lib/connections/nodemailer.js
```diff
@@ -8,21 +8,32 @@ const smtpConfig = {
...
+
+//Do not want the auth object in the smtpConfig object if sendAuthRequest is false, default is true.
```
asahd: Comment should have consistent spacing

```suggestion
// Do not want the auth object in the smtpConfig object if sendAuthRequest is false, default is true.
```

caperneoignis:
Not a problem, will try and fix today.

.env.example
caperneoignis: Can't believe I missed that, forgot to update it.

caperneoignis:
@asahd done, let me know how it looks and if I should make further changes. I also am trying to put all the env variables into the .env.example file so anyone who is curious can see what can actually be changed to clear up any confusion. So if any additional environment variables should be added let me know.

.env.example
caperneoignis: I was thinking the same, but also was trying to avoid long variable names.

PR #1308, opened 2019-01-31
ci(Semantic Release): Automates release process.

ryansmith94:


CHANGELOG.md
```diff
@@ -1,3 +1,5 @@
+# THIS LOG HAS BEEN DEPRECATED AS OF v2.6.4. PLEASE REFER TO OUR RELEASES INSTEAD.
```
asahd: Would be good to have the URL to the releases page here for people who have cloned locally

```suggestion
# THIS LOG HAS BEEN DEPRECATED AS OF v2.6.4. [PLEASE REFER TO OUR RELEASES INSTEAD.](https://github.com/LearningLocker/learninglocker/releases)
```

ht2:
Unfortunately the `after_success_ flag runs after every workflow job, so it's trying to run 3 times. May not be a problem, but probably better we just have a separate job that runs after everything has finished

ryansmith94:
I think I need to make a change to remove the travis condition in our Semantic Release package for this to work too. Alternatively I just switch this to Circle CI.

PR #1310, opened 2019-02-04
feat: Add new route for issuing OAuth2 access_token

h-kanazawa:
https://ht2labs.myjetbrains.com/youtrack/issue/LL-698


api/src/server.js
```diff
@@ -18,6 +18,7 @@ app.use(corsMiddleware);
 app.options('*', corsMiddleware);
 
 app.use(bodyParser.json({ limit: '5mb' }));
+app.use(bodyParser.urlencoded({ extended: false }));
```
ryansmith94: @ht2 do you foresee any issues with this?

cli/src/commands/v2-migrations/20190131000000_oauth_tokens.js
```diff
@@ -0,0 +1,13 @@
+import { getConnection } from 'lib/connections/mongoose';
+
+const up = async () => {
+  const connection = getConnection();
+  await connection.collection('oAuthTokens').createIndex({ expireAt: 1 }, { expireAfterSeconds: 0 });
```
ryansmith94: Nice job, glad you were able to figure this out 👍

api/src/server.js
h-kanazawa: @ht2 for your reference
https://github.com/expressjs/body-parser#bodyparserurlencodedoptions

> Returns middleware that only parses urlencoded bodies and **only looks at requests where the Content-Type header matches the type option.**

> Options
> ...
> **type**
> ... Defaults to `application/x-www-form-urlencoded`

api/src/auth/passport.js
```diff
@@ -224,4 +224,64 @@ if (
...
+      done({ error: 'unsupported_grant_type' });
+      return;
+    }
+
+    if (clientId === undefined || clientSecret === undefined) {
+      done({ error: 'invalid_request' });
+      return;
+    }
+
+    Client.findOne({ 'api.basic_key': clientId }, (err, client) => {
```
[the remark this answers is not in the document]
h-kanazawa: All right. I include `api.basic_key`, `api.basic_secret`, and `isTrusted` checking in the DB call. I remove the error description.

api/src/controllers/AuthController.js
```diff
@@ -214,12 +220,52 @@ const success = (req, res) => {
   res.send('Login success!');
 };
 
+const issueOAuth2AccessToken = (req, res) => {
+  passport.authenticate('OAuth2_Authorization', DEFAULT_PASSPORT_OPTIONS, (err, client) => {
+    if (err) {
+      if (err.isServerError) {
```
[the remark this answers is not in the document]
h-kanazawa: I agree. Default should be server error and Client error should be explicitly declared.

api/src/server.js
ht2: Noted - no can't see any big issues here

ryansmith94: Ok great thanks for checking 👍

api/src/auth/passport.js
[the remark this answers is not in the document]
ryansmith94: Ok cool thanks Hota 👍

ryansmith94:
Note for Asahd: Tested migration worked.

PR #1314, opened 2019-02-12
LL-704 feat: Add contextActivities.definition.type filter

h-kanazawa:
Closes https://github.com/LearningLocker/enterprise/issues/895
https://ht2labs.myjetbrains.com/youtrack/issue/LL-704

Gives ability to filter the contextActivities when being grouping, to ensure we group on only one type of Activity at a time.

E.g. if a bunch of statements all had contextActivities.grouping containing an Organisation, a Course and a Level, I may want to group by the contextActivities.grouping only where the type is Course

## How to review
- Insert a bunch of statements with an assortment of contextActivities, some with a `definition.type` value
- Attempt to group on one particular type (should not include the other types in the grouping output
- Attempt to group with no type (should be existing behaviour)


ht2:
It seems that even when there is data in the cache that should be picked up here:

image

For some reason the drop-down only offers the default option, and does not seem to make an attempt to fetch the data:

image

PR #1318, opened 2019-02-12
findAndModify rewrites

ht2:


lib/connections/mongoose.js
```diff
@@ -19,6 +19,7 @@ const createConnection = () => {
   logger.silly('Creating Mongo connection', dbpath, serverOptions);
   return mongoose.createConnection(dbpath, {
     promiseLibrary: Promise,
+    useFindAndModify: false,
```
ht2: I see no evidence this actually did anything and can probably be killed

ryansmith94: Yeah probably just kill that.

PR #1322, opened 2019-02-20
Moved the persona lock to after validation.

cbishopvelti:
Closes https://github.com/LearningLocker/enterprise/issues/887

## To test
- Create a CSV import with MANY MANY rows of the same INVALID data for an existing persona
  - e.g. Ensure one set of actor data matches an existing persona's identifier
  - But additional idents should be invalid - e.g. a homepage that is an email, or an email column that isn't an email etc
- Spin up multiple workers
- Import and check that you get back validation errors
- Multiple workers should be able to process multiple jobs for the same "persona" simulataneosly
  - Previously, because the persona was locked before we validated the data, it meant only 1 worker could work on any single row of data at a time
  - Now, we validate before locking the persona, meaning we can process multiple rows at once


worker/src/handlers/importPersonas/importPersonaHandler.js
```diff
@@ -98,28 +101,32 @@ export default personaService => async ({
...
+        queueName: PERSONA_IMPORT_QUEUE,
+        payload: {
+          index,
+          data,
+          personaImportId,
+          structure,
+          organisation
+        }
+      });
+      done();
```
ryansmith94: Note to self this done might be called multiple times.

worker/src/handlers/importPersonas/importPersonaHandler.js
```diff
@@ -98,28 +101,32 @@ export default personaService => async ({
...
+      return;
```
ryansmith94: Note to self this now returns out of a different function hence the code can continue past this.

lib/services/importPersonas/importPersona.js
```diff
@@ -71,6 +72,11 @@ export default ({
     };
   }
 
+  // Establish lock here
+  if (await afterValidationCb() === false) {
+    return;
```
ryansmith94: Returning without an object will cause the calling code to throw an error.

worker/src/handlers/importPersonas/importPersonaHandler.js
```diff
@@ -98,28 +101,33 @@ export default personaService => async ({
...
+        }
+      });
+      done();
+      return false;
+    }
+    return true;
+  };
 
   // Do the magic
   const { processedCount, totalCount } = await importPersona({
```
ryansmith94: Currently this will throw an exception because `importPersona` can now return `undefined` instead of an object.

worker/src/handlers/importPersonas/importPersonaHandler.js
```diff
@@ -98,28 +101,33 @@ export default personaService => async ({
...
+        queueName: PERSONA_IMPORT_QUEUE,
+        payload: {
+          index,
+          data,
+          personaImportId,
+          structure,
+          organisation
+        }
+      });
+      done();
```
ryansmith94: If the call to `importPersona` does not throw an error below, this `done` callback will be called again on line 133 in this same file.

PR #1325, opened 2019-02-22
Fixed addCrudFunctions recursively importing.

cbishopvelti:


lib/services/auth/tests/modelFilters/visualisation-test.js
```diff
@@ -1,4 +1,6 @@
 import testShareableModel from 'lib/services/auth/tests/utils/testShareableModel';
+import 'lib/models/dashboard';
```
ryansmith94: What was the reason for this import?

ryansmith94: Chris said this is needed because the tests don't utilise this model elsewhere anymore due to these changes. Therefore this is needed to register the model with Mongoose.

ht2:
Full regression test needed by QA for this, but don't add to QA needed just yet. We'll wait until something else needs a full test too and do them together

PR #1338, opened 2019-03-15
LL-595 Query migration: $or/$nor to $in/$nin

h-kanazawa:

Closes LearningLocker/enterprise#888
Closes https://ht2labs.myjetbrains.com/youtrack/issue/LL-595

- Update queryBuilder
- Add queries migration
  - old queries are stored `old*` field
  - `hasBeenMigrated` field shows whether the entity has been migrated or not
  - The following criteria still use $or/$nor
    - Who > Actor
    - Who > Additional Data > *
    - Where > instructor


cli/src/commands/migrateToInQueries.js
```diff
@@ -0,0 +1,269 @@
...
+  return true;
+};
+
+/**
+ * Convert queries on immutable object layer
+ *
+ * @param {immutable.Map} oldQuery
+ * @returns {immutable.Map}
+ */
+const _buildNewQuery = (oldQuery) => {
```
ryansmith94: I can see that we've exported another function here called `buildNewQuery`, hence why you've put an underscore on it. Do you think there's a better name to give to these two functions so that we don't need the underscore?

cli/src/commands/migrateToInQueries.js
```diff
@@ -0,0 +1,269 @@
...
+      organisationSetting.oldFilter = organisationSetting.filter;
+      organisationSetting.filter = newFilter;
+
+      logger.info('convert organisationSettings filter');
+      logger.info(oldFilter);
+      logger.info(newFilter);
+    });
+
+    user.hasBeenMigrated = true;
+    await user.save();
```
ryansmith94: I suspect @ht2 and @asahd might prefer us to use bulk ops, but given the small size of these collections this looping and awaiting might be ok. What do you think @ht2 and @asahd?

lib/models/dashboard.js
```diff
@@ -28,9 +28,11 @@ const widgetSchema = new mongoose.Schema({
   h: { type: Number, default: 4 }
 });
 
+// TODO: Remove `oldFilter` after we confirm success of $in migration
```
ryansmith94: Thanks for these comments.

ui/src/containers/BasicQueryBuilder/ActorDiscrete/Criteria.js
```diff
@@ -0,0 +1,88 @@
+import React, { Component, PropTypes } from 'react';
```
ryansmith94: Thank you for taking the time to make these additional components and not hacking the existing ones 👍

cli/src/commands/migrateToInQueries.js
ht2: Pretty sure its just to denote its a private method (not exported) - not a big one for me

cli/src/commands/migrateToInQueries.js
ht2: Should be fine, only a few hundred of these to worry about and does mean each one is done atomically

ryansmith94: Okay nice thanks, yeah agreed 👍

cli/src/commands/migrateToInQueries.js
ryansmith94: Yeah just popped out to me, not that fussed either.

h-kanazawa:
fixed issued QA team found.

- Causes
  - Issue 01: forgot `await` in migrate script. fixed at  https://github.com/LearningLocker/learninglocker/pull/1338/commits/7a7f072f6e5c7bdc065aaea9c6a4416de7a94b66
  - Issue 02: forgot migrating `queries` collection. fixed at https://github.com/LearningLocker/learninglocker/pull/1338/commits/81f6d8fc5edeb27bfd4f1b9be7e19bca7f309999
  - Issue 03: did not consider statement filtering in Source page. fixed at https://github.com/LearningLocker/learninglocker/pull/1338/commits/4832dde69fd1cc50a1250423c7420ed1760c6647 and https://github.com/LearningLocker/learninglocker/pull/1338/commits/001199b6cc27ca458914e05e1574542ab9e23604

@asahd

asahd:
Thank you @h-kanazawa!

asahd:
@h-kanazawa 🎉 👍

Issue #1339, opened 2019-03-17
Heroku support on frontend applications


SindreSvendby:
#### Is your feature request related to a problem? Please describe.
I can't run learning locker with Heroku since the application does not bind to $PORT env var 
that Heroku is using 

#### Describe the solution you'd like
Right now the different parts of the application is using env var to choose which port to connect to.
[Configuration](http://docs.learninglocker.net/guides-configuring/) mentions `UI_PORT` for the frontend and `EXPRESS_PORT` for the [xapi-service](https://github.com/LearningLocker/xapi-service)

#### Describe alternatives you've considered
1. I checked if it was possible to say to Heroku which env var it could use to export the port to bind to except `PORT`, but did not find a way to do that.

2. I googled around and tested to see if I was able to use one env var in another, as in setting `EXPRESS_PORT=$PORT`, but was not able to do anything like this. 

I have [created this as a PR](https://github.com/LearningLocker/xapi-service/pull/272) for the `xapi-service`, but the issue is in this repo, 
since I would like to do the same for the UI, and since the `xapi-service` had disabled issues.

PR #1340, opened 2019-03-17
Support PORT in addition to UI_PORT

SindreSvendby:
Closes part 2 of  https://github.com/LearningLocker/learninglocker/issues/1339


ryansmith94:
Hi @SindreSvendby, thanks for the pull request and issue. I'll try to take a look at these today.

SindreSvendby:
Happy to discuss this. 

Could you explain a bit why you think it needs to be behind nginx?

is there anything in the xapi that needs the learninglocker app and expects to find it on the same domain? 

Is there anything in the learninglocker that needs the xapi and expects to find it on the same domain? 

Is there any caching or other stuff done by nginx so any of the service do not work without nginx on top of it?

ht2:
##### is there anything in the xapi that needs the learninglocker app and expects to find it on the same domain?

No, the xAPI can run as it's own service as long as it connect to the same database

##### Is there anything in the learninglocker that needs the xapi and expects to find it on the same domain?
Technically yes - the UI presents the xAPI endpoint to the end user on the Clients page based on the SITE_URL in the webapp's `.env`

##### Is there any caching or other stuff done by nginx so any of the service do not work without nginx on top of it?

The UI (typically running on port 3000 for example) is configured to proxy any static media and will also proxy to the API (for routes prefixed as such). However as typically only one UI process is ever run, all API requests are throttled through that one process, hence putting a reverse proxy in front of the API separates the concerns so that the UI does not need to handle that load.


All in all, I don't object to this PR, but you should know the limitations and reasons why putting it behind nginx (or any reverse proxy for that matter) is probably the better solution. The initial push back was as we would be worried that introducing this method would create a precedent for other users to follow that we do not believe is in the best interests of them.

I'm happy to accept the PR, but we will probably leave this undocumented for the time being - or if we do add it place reference to this PR (and this very comment!)

SindreSvendby:
Thanks for the detailed feedback. 

For me this make sense, and I think some users would like to be able to offload all the infrastructure to Heroku or similar, so they can use it like a PaaS. 

I have added a [comment](https://github.com/LearningLocker/xapi-service/pull/272#discussion_r268339614) on the other PR. But except from that I'm think this is ready to go. 

Please let me know if there is any additional work required from my side. 

Thanks again for a great product! :+1:  :hugs:

ryansmith94:
Hey @SindreSvendby are you happy with me committing those suggestions in the xapi-service? I'll get this merged after 👍

SindreSvendby:
Sorry for the delay @ryansmith94, I was on sick leave. Your suggestions are applied now!

ryansmith94:
No worries @SindreSvendby, thank you, just waiting for CircleCI to pick up your two PRs to test them.

PR #1344, opened 2019-03-27
Added deletion of statements.

cbishopvelti:

Closes LearningLocker/enterprise#922.


api/src/routes/HttpRoutes.js
```diff
@@ -265,7 +270,20 @@ restify.serve(router, Dashboard);
 restify.serve(router, LRS);
 restify.serve(router, Statement, {
   preCreate: (req, res) => res.sendStatus(405),
-  preDelete: (req, res) => res.sendStatus(405),
+  preDelete: (req, res, next) => {
+    const authInfo = getAuthFromRequest(req);
+    if (!get(process.env, 'ENABLE_SINGLE_STATEMENT_DELETION', true)) {
```
ryansmith94: If this was set to `false` in the ".env" file this value would be truthy and hence incorrectly allow the deletion of a statement. I suggest using the boolean package to fix this issue since it will parse a "false" string to a falsy value.

lib/constants/scopes.js
```diff
@@ -18,17 +20,29 @@ export const XAPI_STATEMENTS_WRITE = 'statements/write';
...
   [XAPI_STATEMENTS_READ]: 'Read all statements',
   [XAPI_STATEMENTS_WRITE]: 'Write statements (must be used with a read scope)',
   [XAPI_STATEMENTS_READ_MINE]: 'Read my statements',
+  [XAPI_STATEMENTS_DELETE]: 'Delete statements (not covered by all scope)',
   [XAPI_STATE_ALL]: 'Access state',
   [XAPI_PROFILE_ALL]: 'Access profiles',
 };
 
+export const getXapiScopes = (
+  deleteStatementsEnabled = get(process.env, 'ENABLE_SINGLE_STATEMENT_DELETION', true)
```
ryansmith94: I can't remember what webpack does in this case, but does this potentially expose our entire `process.env` since I think webpack replaces references to the `process.env`. If this is the case we need explicitly look for `process.env. ENABLE_SINGLE_STATEMENT_DELETION`.

cbishopvelti: Not sure how that would work? There is no entry point for dotenv config in the client context.

ryansmith94: @ht2 any idea whether Webpack injects this?

ht2: I'm not sure guys, is this used client side, and does it pull the correct value from the .env when you've tested it? Presumably this is built and read from the .env when the client server is started, but Chris is probably right that there isn't an entry point

ryansmith94: I think we just need to log out the `process.env` then and see what happens.

ryansmith94: Ok I think this is fine, I've added the var to the webpack build.

api/src/routes/HttpRoutes.js
```diff
@@ -265,7 +270,20 @@ restify.serve(router, Dashboard);
 restify.serve(router, LRS);
 restify.serve(router, Statement, {
   preCreate: (req, res) => res.sendStatus(405),
-  preDelete: (req, res) => res.sendStatus(405),
+  preDelete: (req, res, next) => {
+    const authInfo = getAuthFromRequest(req);
```
ryansmith94: Note to self: this might allow the deletion of multiple statements.

cbishopvelti:
This branch has been merged into batch deletion branch issue-LL-902. So this branch is now deprecated.

PR #1345, opened 2019-03-28
fix(Workers): Add better error logging to SQS failures

ht2:
n/a

Currently the SQS worker outputs `[ERROR] SQS HANDLE JOB` plus the JS error. Would be much more useful to see which queue and payload led to the error


ryansmith94:
Approved, but obviously the build doesn't approve lol. You need to just remove the payload from that second log.

ht2:
Woops - I actually wanted the payload for that, it just wasn't in a var

PR #1346, opened 2019-04-04
feat: Allows deletion of statements (single and batch).

cbishopvelti:

Closes LearningLocker/enterprise#901 (Batch deletions)
Closes LearningLocker/enterprise#922 (Single deletion)


ui/src/controllers/renderApp.js
```diff
@@ -23,7 +23,8 @@ export default async (req, res, next) => {
 
     data.state = {
       app: {
-        RESTRICT_CREATE_ORGANISATION: boolean(defaultTo(process.env.RESTRICT_CREATE_ORGANISATION, true))
+        RESTRICT_CREATE_ORGANISATION: boolean(defaultTo(process.env.RESTRICT_CREATE_ORGANISATION, true)),
+        ENABLE_SINGLE_STATEMENT_DELETION: boolean(defaultTo(process.env.ENABLE_SINGLE_STATEMENT_DELETION, true))
```
ryansmith94: Shouldn't there be something in here for batch deletions too @cbishopvelti?

api/src/controllers/BatchDeleteController.js
```diff
@@ -0,0 +1,53 @@
...
+  console.log('authInfo', JSON.stringify(authInfo, null, 2));
+  const batchDelete = new BatchDelete({
+    organisation: getOrgFromAuthInfo(authInfo),
+    total,
+    filter: req.query.filter,
+    client: get(authInfo, 'client.id')
+  });
+  await batchDelete.save();
+
+  await publish({
```
ryansmith94: I'm not sure what this publishes to a worker, can you clarify please @cbishopvelti?

ui/src/containers/ClientForm/__tests__/ClientForm.spec.js
```diff
@@ -11,7 +11,12 @@ import AuthorityEditor from '../AuthorityEditor';
 global.__DEVELOPMENT__ = true;
 global.__CLIENT__ = false;
 const createMockStore = configureMockStore([thunkMiddleware]);
-const store = createMockStore({ models: new Map(), metadata: new Map(), pagination: new Map() });
+const store = createMockStore({
+  app: new Map({ ENABLE_SINGLE_STATEMENT_DELETION: false }),
```
ryansmith94: Should be something here for batches too @cbishopvelti?

worker/src/handlers/batchStatementDeletion/batchStatementDeletion.js
```diff
@@ -0,0 +1,125 @@
...
+    }
+    throw err;
+  }
+
+  const filter = {
+    ...(await parseQuery(batchDelete.filter)),
+    ...scopeFilter
+  };
+
+  const docs = await Statement.aggregate()
```
ryansmith94: Why did you use `aggregate` instead of `find` here @cbishopvelti?

ui/src/controllers/renderApp.js
ryansmith94: One scope for deletion statements, however, this wouldn't be visible unless single statement deletion is enabled.

worker/src/handlers/batchStatementDeletion/batchStatementDeletion.js
ryansmith94: @ht2 Chris thinks the aggregate performs better than the find. He had also thought that the find didn't take a projection, but I've shown him that [it does take a find in the docs](https://mongoosejs.com/docs/api.html#model_Model.find).

api/src/controllers/BatchDeleteController.js
ryansmith94: Chris said that the scheduler will trigger a worker job to delete a batch and the worker job will delete a batch then spawn a subsequent job for the next batch deletion if more deletions are required and we're within the time window for allowing the deletion. Note that the locking in this PR is to ensure the scheduler doesn't run on multiple servers, that locking also exists in other schedulers for the same reason.

worker/src/handlers/batchStatementDeletion/batchStatementDeletion.js
ht2: See here for a good example where we use a cursor on a filtered query with a lean projection of just _id: https://github.com/LearningLocker/enterprise/blob/master/lib/services/journey/getJourneyStatements.js#L18

api/src/controllers/BatchDeleteController.js
ht2: I think this makes sense - we don't want one process to handle the deletion solely as it would mean the job would be lost if the server fell over etc. 

The locking however doesn't feel right, as in flight queues are only pulled by one worker at a time, so that should be fine. 

Instead, where locking _is_ needed is when the scheduler locks using _Redis_ to ensure no two schedulers can run the scheduled job at the same time. See here: https://github.com/LearningLocker/learninglocker/blob/master/cli/src/scheduler.js#L42

ryansmith94: Yeah locking is only for the scheduler here.

ui/src/controllers/renderApp.js
ht2: OK, happy for there to only be one scope, but in which case we should rename the env variable accordingly and also remove the additional env that _does_ reference batch deletions!

https://github.com/LearningLocker/learninglocker/pull/1346/files/df8441e126271e968fe32932a78411d9d9d4ac52#diff-cbfd64a28982a1818f2b5f16e7f9d634R218

ryansmith94: Ok thanks James 👍

worker/src/handlers/batchStatementDeletion/batchStatementDeletion.js
ryansmith94: Ok thank you 👍

ht2:
YUUUUUUUUUUUUUUUUUUSSSSSSSSSSSSSSSSS

api/src/controllers/BatchDeleteController.js
```diff
@@ -1,39 +1,61 @@
...
+import { get, isEmpty } from 'lodash';
+import mongoose from 'mongoose';
 
+const objectId = mongoose.Types.ObjectId;
 
-const initialiseBatchDelete = catchErrors(async (req, res) => {
-  // Authenticate
+const checkDeletionsEnabled = () => {
+  if (boolean(get(process.env, 'ENABLE_STATEMENT_DELETION', true)) === false) {
+    // Clear the job and don't do anything
```
ryansmith94: Delete this line

api/src/routes/tests/batchDeleteController/batchDeletion-intiailise-test.js
```diff
@@ -0,0 +1,60 @@
+
```
ryansmith94: Delete blank lines

api/src/routes/tests/batchDeleteController/batchDeletion-terminate-test.js
```diff
@@ -0,0 +1,84 @@
+
```
ryansmith94: Delete blank lines

api/src/routes/tests/batchDeleteController/batchDeletion-terminateAll-test.js
```diff
@@ -0,0 +1,77 @@
+
```
ryansmith94: Delete blank lines.

worker/src/handlers/batchStatementDeletion/batchStatementDeletion.js
```diff
@@ -94,6 +112,7 @@ export default async ({
   }, {});
 
   let doneUpdate = {};
+  console.log(result, filter);
```
ryansmith94: Delete log

Issue #1291, opened 2019-01-03
$sample as start in mongo pipeline not respected as start


davidavdav:
**Version**
{{branch}} at f4fd50090b44b0b37fcd23c5f664619ee5c0e279 on 2.6.1

**Steps to reproduce the bug**
1. use database with large statements collection
2. use aggregate request starting with { $sample: size: 10 }
3. use mongo shell `db.currentOp()` to see what takes so long
4. observe that in the long lasting operation, the pipeline has been prepended with matches to `lrs_id` and `organisation`

**Expected behaviour**
The $sample pipeline statement, when used as a first, [should stay first](https://docs.mongodb.com/manual/reference/operator/aggregation/sample/#pipe._S_sample), otherwise mongo will do a full document scan---and it should sample instead. 

**Actual behaviour**
$sample is moved to 3rd place in the pipeline, making it do a full scan, which takes forever on large collections.  

**Additional information**
OS: linux
our collection has about 7 million statements at the moment.

davidavdav:
I have a fix for this now, I'll try to make a PR soon

ryansmith94:
PR #1292 was closed due to inactivity. We need to implement a `sampleSize` query parameter to allow this without modifying the query in a new PR.

ryansmith94:
Resolved by [v3.7.0](https://github.com/LearningLocker/learninglocker/releases/tag/v3.7.0) 👍

PR #1347, opened 2019-04-06
Add option sampleSize to efficiently sample statements, fixes 1291

davidavdav:
As discussed in #1291 introducing a `$sample` at the start of the pipeline of an aggregate query can take very long when there are many statements.  This is because learning locker inserts two matches at the head of the pipeline, so that the user-issued `$sample` is third.  This makes mongo do a full document scan, instead of taking a more efficient strategy. 

In #1292 a solution was suggested which keeps `$sample` statements first, but a preference was expressed for addressing this issue using an option. 

This PR allows a parameter `sampleSize` to be set as an option (default `-1`, meaning disabled) that causes a `{ $sample: {size: sampleSize } }` to be placed in the pipeline ahead of anything else, allowing efficient sampling by mongo. 

The user can still use a `$sample` statement further down the pipeline, and of course also combine with a `sampleSize` option.


api/src/controllers/StatementController.js
```diff
@@ -16,14 +16,16 @@ const aggregate = (req) => {
   const maxTimeMS = Number(req.query.maxTimeMS) || MAX_TIME_MS;
   const maxScan = Number(req.query.maxScan) || MAX_SCAN;
   const pipeline = JSON.parse(req.query.pipeline);
+  const sampleSize = Number(req.query.sampleSize) || -1;
```
ryansmith94: I think I would prefer to default this to undefined since we don't usually allow error-like values like this in Learning Locker's code.

davidavdav: OK, changed it to `undefined`.

ryansmith94: Sweet thank you! 👍

lib/models/statement.js
```diff
@@ -187,6 +188,10 @@ schema.statics.aggregateByAuth = function aggregateByAuth(
     if (authInfo.client && authInfo.client.lrs_id) {
       parsedPipeline.unshift({ $match: { lrs_id: objectId(authInfo.client.lrs_id) } });
     }
+
+    if (sampleSize) {
```
ryansmith94:
```suggestion
    if (sampleSize !== undefined && sampleSize > 0) {
```

ryansmith94: @davidavdav this has been through QA, can you please just commit this suggestion so that this condition is a little clearer? Then we're happy to merge and release this 👍

davidavdav: Strange... in the email from github the suggestion is 
```
if (sampleSize !== undefined && sampleSize > 0)
```
whereas in the github PR-page the suggestions has been reduced to 
```
if (sampleSize > 0)
```
I suppose `undefined > 0 === false`, anyway, so I'll click on the "Commit suggestion" button and see how that works...

ryansmith94: Yeah sorry I changed the suggestion after realising `undefined > 0` would be `false`.

ryansmith94: Thanks for committing so quickly. I'll wait for the build to pass then merge.

PR #1359, opened 2019-04-26
Issue 937 - $lookup

cbishopvelti:

Closes https://github.com/LearningLocker/enterprise/issues/937.


lib/helpers/filter$lookup.js
```diff
@@ -0,0 +1,66 @@
...
+const ALLOWED_LOOKUPS = [
+  'personas',
+  'personaIdentifiers',
+  'personaAttributes',
+  'journeys',
+  'journeyprogresses',
+  'statements',
+];
+
+const shouldRemove = (value, key) => {
```
ryansmith94: Since this looks similar to [remove$out](https://github.com/LearningLocker/learninglocker/blob/master/lib/helpers/remove$out.js), could this be simplified to the code below?

```js
const shouldRemove = (value, key) => {
  return (
    _.has(value, ['$lookup', 'from') &&
    includes(ALLOWED_LOOKUPS, get(value, ['$lookup', 'from']))
  );
}
```

lib/helpers/filter$lookup.js
```diff
@@ -0,0 +1,45 @@
...
+  'personaIdentifiers',
+  'personaAttributes',
+  'journeys',
+  'journeyprogresses',
+  'statements',
+];
+
+const shouldRemove = value => (
+    has(value, ['$lookup', 'from']) &&
+    includes(ALLOWED_LOOKUPS, get(value, ['$lookup', 'from']))
```
asahd:
```suggestion
    !includes(ALLOWED_LOOKUPS, get(value, ['$lookup', 'from']))
```

ht2:
Might need a new index on persona attributes $lookup due to organisation being added to $match stage

PR #1363, opened 2019-05-01
Issue 936 - API selects

h-kanazawa:

Closes https://github.com/LearningLocker/enterprise/issues/936.

### DIFF
--- a/api/src/auth/passport.js
+++ b/api/src/auth/passport.js
@@ -135,7 +135,6 @@ passport.use(
       reason: null,
       success: false
     };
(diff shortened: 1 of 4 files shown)


api/src/auth/passport.js
```diff
@@ -135,6 +135,7 @@ passport.use(
       reason: null,
       success: false
     };
+    console.log(`--- passport.js ${password}`);
```
ryansmith94: Probably should remove this line
```suggestion
```

PR #1370, opened 2019-05-09
fix(JWTAuthentication): unexpected 500 responses

h-kanazawa:

How to simulate requests with expired JWT

- Set a short time, e.g. '10s', to `expiresIn` in api/src/auth
  - https://github.com/LearningLocker/learninglocker/blob/12892111d74a27c473891b36b2b8eef2926ba788/api/src/auth/jwt.js#L52
- Login and wait some seconds, and move pages. It will dispatch API Request to /api/connection
- The responses code of /api/connection should be 401, instead of 500, which is before fixing.


ht2:
https://github.com/LearningLocker/enterprise/issues/940

PR #1376, opened 2019-05-14
fix(Statement count): Await save

ht2:
@ryansmith94 I think this await was missing - would you concur? It deff skipped through the save for me locally/


ryansmith94:
Seems extremely likely @ht2 👍

PR #1377, opened 2019-05-14
refresh access token

h-kanazawa:
Close https://github.com/LearningLocker/enterprise/issues/940
Close https://github.com/LearningLocker/enterprise/issues/967


api/src/auth/jwt.js
```diff
@@ -120,8 +169,12 @@ export {
...
   createUserJWT,
+  createUserRefreshTokenPayload,
+  createUserRefreshJWT,
   createOrgTokenPayload,
   createOrgJWT,
+  createOrgRefreshTokenPayload,
+  createOrgRefreshJWT,
   createDashboardTokenPayload,
-  createDashboardJWT
+  createDashboardJWT,
```
ryansmith94: Do we need to do this for dashboard tokens too @ht2?

api/src/controllers/AuthController.js
```diff
@@ -1,17 +1,35 @@
...
+import Unauthorized from 'lib/errors/Unauthorised';
+import { createOrgJWT, createOrgRefreshJWT, createUserJWT, createUserRefreshJWT } from 'api/auth/jwt';
 import { AUTH_FAILURE } from 'api/auth/utils';
 
+const buildRefreshCookieOption = protocol => {
+
+  const c = {
+    path: `/api${AUTH_JWT_REFRESH}`,
+    expires: new Date(Date.now() + 7 * 24 * 60 * 60 * 1000),
+    maxAge: 7 * 24 * 60 * 60 * 1000,
```
ryansmith94: Looks like `7 * 24 * 60 * 60 * 1000` could be a constant, perhaps we call it `refreshExpiration`. What do you think @h-kanazawa?

api/src/controllers/AuthController.js
```diff
@@ -1,17 +1,35 @@
...
   DEFAULT_PASSPORT_OPTIONS
 } from 'lib/constants/auth';
-import { createOrgJWT, createUserJWT } from 'api/auth/jwt';
+import Unauthorized from 'lib/errors/Unauthorised';
+import { createOrgJWT, createOrgRefreshJWT, createUserJWT, createUserRefreshJWT } from 'api/auth/jwt';
 import { AUTH_FAILURE } from 'api/auth/utils';
 
+const buildRefreshCookieOption = protocol => {
+
+  const c = {
```
ryansmith94: We should probably rename `c` to something more meaningful.

api/src/controllers/AuthController.js
```diff
@@ -161,13 +179,51 @@ const jwt = (req, res, next) => {
     user.authFailedAttempts = 0;
     user.authLastAttempt = new Date();
     return user.save(() =>
-       createUserJWT(user).then()
-          .then(token => res.send(token))
-          .catch(authFailure)
+      Promise.all([createUserJWT(user), createUserRefreshJWT(user)])
```
ryansmith94: Good use of Promise.all to do this in parallel.

api/src/controllers/AuthController.js
```diff
@@ -161,13 +179,51 @@ const jwt = (req, res, next) => {
...
+      } else {
+        throw new Unauthorized('Invalid tokenType');
+      }
+    } catch (err) {
+      if (!['JsonWebTokenError', 'TokenExpiredError'].includes(err.name) && !(err instanceof Unauthorized)) {
+        console.error(err);
+      }
+      res.status(401).send('Unauthorized');
+    }
+  })().catch(next);
```
ryansmith94: I'm a little uncertain about what happens when you call next with an error like Unauthorised. Could you please clarify what happens in these cases @h-kanazawa? Typically we would just handle all errors in the route handler to avoid having to read Express docs and trace down any code that might handle this exception elsewhere.

api/src/routes/tests/http-test.js
```diff
@@ -47,6 +50,246 @@ describe('API HTTP Route tests', function describeTest() {
...
+      // To confirm `iat` and `exp` are changed
+      await delay(1000);
+      refreshTokenCookie = refreshTokenCookie.split('; ')[0];
+
+      await apiApp
+        .post(routes.AUTH_JWT_REFRESH)
+        .set('Cookie', refreshTokenCookie)
+        .send({ id: '561a679c0c5d017e4004714f', tokenType: 'user' })
+        .expect(200)
+        .expect('Content-Type', /text\/html/)
```
ryansmith94: I think it might be odd to return `text/html` as the content type because the response is actually just the token as text rather than html. What do you think @h-kanazawa?

api/src/auth/jwt.js
ht2: Good catch - very much depends if the dashboard tokens have an expiry. Given they are public and can be regenerated simply by refreshing the page, we probably don't need to; however some of the logic in this PR may change that behaviour, so it should be double checked.

api/src/controllers/AuthController.js
ht2: Can we make it an env value to with that as the default?

api/src/controllers/AuthController.js
ht2: I think we'd be back to an unhandled exception with a 500 thrown by Express. Might be nicer to wrap in [`catchErrors`](https://github.com/LearningLocker/learninglocker/blob/a3e3ab4290f22d579f6e908b38231435b4a57bf4/api/src/controllers/ExportController.js#L13) as we do elsewhere, but unsure if that presents issues

api/src/routes/tests/http-test.js
ht2: Perhaps `text/plain`?

ht2: Wait, isn't it `application/json`? I'm guessing that Content-Type is there for a reason though....

api/src/auth/jwt.js
ryansmith94: Yeah ok, @h-kanazawa can you please double check this when you go through these comments?

api/src/controllers/AuthController.js
ryansmith94: Would you ever consider changing it? I'd rather avoid having too many env values if possible.

api/src/routes/tests/http-test.js
ryansmith94: Either `application/json` or `text/plain` would be fine I think. Yeah I'm unsure why it's `text/html` because I think it's just text.

api/src/controllers/AuthController.js
ht2: I guess the refresh is less important, just thinking for QA but they shouldn't need to fiddle with this number

ryansmith94: Ah right yeah, can understand the need for it in QA. Your call on whether this is required @ht2.

api/src/auth/jwt.js
h-kanazawa: I changed dashboard refresh token's expiry, but the other behaviours are not changed. Dashboard tokens are not refreshed.

ryansmith94: Ok thanks @h-kanazawa 👍

api/src/routes/tests/http-test.js
h-kanazawa: I add 3 env variables

- `REFRESH_TOKEN_EXPIRATION_MSEC` (for cookie's expires and maxAge) e.g. "3600000"
- `REFRESH_TOKEN_EXPIRATION`(for jwt exp) e.g. "7d" This value is expected to be same length with `REFRESH_TOKEN_EXPIRATION_MSEC`
- `ACCESS_TOKEN_EXPIRATION` (for jwt exp) e.g. "1h"

api/src/controllers/AuthController.js
h-kanazawa: I add 3 env variables for QA

- `REFRESH_TOKEN_EXPIRATION_MSEC` (for cookie's expires and maxAge) e.g. "3600000"
- `REFRESH_TOKEN_EXPIRATION`(for jwt exp) e.g. "7d" This value is expected to be same length with `REFRESH_TOKEN_EXPIRATION_MSEC`
- `ACCESS_TOKEN_EXPIRATION` (for jwt exp) e.g. "1h"

ryansmith94: Okay thanks Hota 👍

api/src/controllers/AuthController.js
h-kanazawa: Sorry. This async anonymous function ignores errors. So `.catch` will never be called. Furthermore, `jwtRefresh` doesn't need `next` function. I rewrite this with catchErrors.

api/src/routes/tests/http-test.js
h-kanazawa: I changed it to text/plain.

PR #900, opened 2017-02-20
Fixes for issue 895 (and additional error found)

davidpesce:
Fixes for #895. First commit had IDE formatting issues. Full PR should resolve that.

Previous code returned a success message regardless of the EloquentUserRepository.php verifyEmail function. This should be resolved now.

PR #1380, opened 2019-05-28
Set visibilityTimeout.

cbishopvelti:

Closes https://github.com/LearningLocker/enterprise/issues/900.


ryansmith94:
Assigning @ht2 to dev review because I have quite a limited understanding of SQS right now unfortunately whereas James understands it really well.

ht2:
Think we're blocked on this due to SQS throwing its toys out the pram whenever the timeout changes on a queue.

Somewhat related to #900

Basically, instead of setting the visibilityTimeout on queue creation, we should update the attribute after creation to ensure we don't get collisions.

ht2:
https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_SetQueueAttributes.html

ht2:
Closing this, as @cbishopvelti says that this is now handled in #1397

PR #1392, opened 2019-06-13
fix(QueryBuilderCacheValues): enable to overwrite displayed object name

h-kanazawa:
Closes https://github.com/LearningLocker/enterprise/issues/968

I described replication method and test results on [the issue](https://github.com/LearningLocker/enterprise/issues/968)


lib/services/querybuildercache/saveCacheValues.js
```diff
@@ -12,17 +12,27 @@ export default function (caches, organisation, batch) {
...
-      updatedAt
-    });
+
+    // combination of organisation, path, and hash is unique key.
+    // Overwrite
+    batch
+      .find({
+        organisation,
+        path: path.join('.'),
+        hash,
```
ryansmith94: Is the object name used as part of the hash @h-kanazawa?

h-kanazawa: I think `hash` is built from only `value`.

We want queryBuilderCacheValues to be unique on the combination of  `organisation`, `path`, and `value`.
Because when we fetch queryBuilderCacheValues via API */api/connection/querybuildercachevalue*, we use `organisation` and `path` as filter. And "What" dropdown uses the `value` as it key.

I guess that we convert `value` to `hash` because it is an object.

If `hash` is created from `value` and the object name (`display`), queryBuilderCacheValues will have duplicated  the combination of  `organisation`, `path`, and `value`. That is not expected.

Is this an answer for your question, @ryansmith94 ?

ryansmith94: I believe hash is built from the value and yeah we convert the value to the hash because it's an object. I'm concerned that we might be hashing the entire object with the display property included. In which case, we'd first create a QBCV (query builder cache value) using Object A as the hash, but then we'd try to find the QBCV for Object A by hashing Object B which would not match Object A. Therefore, I believe the change in this PR may result in a new QBCV everytime the display name changes, rather than updating the existing QBCV.

#### Object A
```json
{
  "id": "https://ht2labs.com/Question10",
  "objectType": "Activity",
  "definition": {
    "type": "https://ht2labs.com/types/Q",
    "name": {
      "en-GB": "Question 10"
    }
  }
}
```

#### Object B
```json
{
  "id": "https://ht2labs.com/Question10",
  "objectType": "Activity",
  "definition": {
    "type": "https://ht2labs.com/types/Q",
    "name": {
      "en-GB": "Answer 10"
    }
  }
}
```

h-kanazawa: Which name do we use for option label if we store multiple QBCVs to `id` ?

```
<Dropdown index={selctedIndex} >
  <OptionItem index="https://ht2labs.com/Question10" label="Question 10" />
  <OptionItem index="https://ht2labs.com/Question10" label="Answer 10" />
</Dropdown>
```

This is a simplified dropdown model. The dropdown shows an option matching on their index. So `id` must be unique in the options.

ryansmith94: We should only display one option and we should display the most recent. However, we should ideally solve this problem by updating the correct the QBCV in the database so that we only return one from the API.

h-kanazawa: Yes.

ryansmith94: Are you able to confirm locally if sending those two objects in statements above create two QBCVs or one QBCV?

h-kanazawa: One QBCV. A newer ("Answer 10") overwrites the QBCV document. "Question 10" is gone.

PR #1395, opened 2019-06-17
Support first name and last name in Personas Import

h-kanazawa:
Close https://github.com/LearningLocker/enterprise/issues/980


lib/services/importPersonas/personasImportHelpers.js
```diff
@@ -137,9 +139,35 @@ export const getPersonaName = ({
...
+  const firstNameColumn = chain(structure)
+    .findKey(item => item.columnType === COLUMN_FIRST_NAME)
+    .value();
+
+  const lastNameColumn = chain(structure)
+    .findKey(item => item.columnType === COLUMN_LAST_NAME)
+    .value();
 
-  return row[nameColumn];
+  const firstName = row[firstNameColumn];
```
cbishopvelti: trim(`${get(row, firstNameColumn)} ${get(row, lastNameColumn)}`);

Maybe a little more concise.

lib/services/queue/bull/index.js
```diff
@@ -99,9 +99,9 @@ export const subscribe = ({
     }
 
     try {
-      queue.process((job, jobDone) =>
-        handler(job.data, jobDone)
-      );
+      queue.process((job, jobDone) => {
```
cbishopvelti: This wont lint.

lib/services/queue/bull/index.js
```diff
@@ -110,10 +110,12 @@ export const subscribe = ({
...
 };
 
-export const unsubscribeAll = () => {
+export const unsubscribeAll = async () => {
   const queues = map(Bull.queues, queue => queue);
-  forEach(queues, ((queue) => {
-    queue.close(true);
-  }));
+
+  for (const queue of queues) {
```
cbishopvelti: Avoid use of for loop. Use bluebird: http://bluebirdjs.com/docs/api-reference.html. Promise.map maybe.

lib/services/queue/bull/index.js
h-kanazawa: Yes, I'll fix this.

lib/services/importPersonas/personasImportHelpers.js
h-kanazawa: Nice.
If a first name starts with spaces, e.g. "  abc", we should leave them, shouldn't we?
@cbishopvelti @ryansmith94

h-kanazawa: If trimming is ok, I'll refactor this function.

ryansmith94: I don't have any issues with trim except for the inconsistency.

Let's say one row has a first name of "a " and no last name, their full name would be "a". However, another row could have the same first name with a last name of "b" and their full name would be "a  b" (two spaces between). You could solve this issue with ```trim(`${trim(firstName)} ${trim(lastName)}`)```.

Note that this returns an empty string when no firstName or lastName are provided, so perhaps we should check for that and return undefined in that case to maintain the current functionality.

h-kanazawa: OK

h-kanazawa:
@cbishopvelti I added trimming process. Can you check that again and move to Ops Review if it's ok.

lib/models/personasImport-test.js
```diff
@@ -180,31 +185,198 @@ describe('personasImport model', () => {
...
+          test2: {
+            columnType: COLUMN_NAME,
+          },
+        };
+
+        const result = validateFullNameAndFirstOrLast(structure);
+        expect(result).to.equal(true);
+      });
+
+      it('should return false if input has first name and full name', () => {
```
ryansmith94: We probably could have just defaulted to using the full name in these last two cases, but this is very explicit and I think I prefer this approach. Thanks for adding tests that were clear to understand 👍

h-kanazawa:
> Importing the persona with changed first name and last name fields: Merges but uses first name only instead of first name + last name.

I can not reproduce this.

1st import
```
_id,firstName,lastName
mailto:u1@example.com,F,L
```

u1@example.com 's name is "F L"

2nd import
```
_id,firstName,lastName
mailto:u1@example.com,UF,UL
```

u1@example.com 's name is "UF UL"

> Importing an updated persona with changed first name, last name and full name fields: Merges but uses first name + last name. In other cases full name field takes priority.

Basically, "first name + last name" and "full name" can not be selected at the same time. So that line should be error.

HT2-QA:
Have discovered the cause of "incorrect" behaviour. This does not seem to be an issue with this PR but how the import UI handles errors.

E.g.
1. Import with Identifier mbox set, all other columns set to none.
2. Correctly set **First name** and **Last name**
3. Set a **Full name** column
4. Press import.
5. Check columns, **Full name**, **First name** and **Last name** show as imported.
6. Refresh page.
7. **Full name** will have changed to **Nothing**.
8. Check new personas, these will show with **First name** and **Last name**

We will make a separate issue for the import UI error handling.

Issue #1394, opened 2019-06-17
Incorrect dashboard URL throws error


ryansmith94:
PR: https://github.com/LearningLocker/learninglocker/pull/1396

#### What version were you using?
v2.4.0

#### What steps can we follow to reproduce the behaviour?
Go to https://lrs-url.com/dashboards/some_incorrect_dashboard_id

#### What is the actual behaviour?
UI server throws uncaught error causing UI server to restart.

#### What is the expected behavior?
UI server should catch the error and return a 404.

h-kanazawa:
@cbishopvelti I could not replicate the uncaught error causing UI server to restart on v2.4.0 and the newest version.

UIServer caught an error and it logged `err` as below. 

```
18|UIServer  | 2019-06-18 14:07:04:049 - error: Error in request {"message":"Cast to ObjectId failed for value \"some_incorrect_dashboard_id\" at path \"_id\" for model \"Dashboard\"","name":"CastError","stringValue":"\"some_incorrect_dashboard_id\"","kind":"ObjectId","value":"some_incorrect_dashboard_id","path":"_id"}
```

Did you replicate the "restart"?

miruser:
Hi LearningLocker.
You forgot to reference the CVE-2019-12834. Can you add it to your release notes, please?
https://nvd.nist.gov/vuln/detail/CVE-2019-12834

ryansmith94:
Hi @miruser, it wasn't forgotten, our policy has been not to immediately detail security patches in our release notes to protect our users that can't upgrade quickly due to their procedures. Since it's been 3 months now, I'll add it to the release notes.

PR #1396, opened 2019-06-18
Fixed UIServer crashing if invalid id is given.

cbishopvelti:

Closes https://github.com/LearningLocker/learninglocker/issues/1394.


h-kanazawa:
LGTM

ui/src/controllers/renderDashboard.js
```diff
@@ -87,7 +87,7 @@ export default (req, res) => {
     })
   )
   .catch((err) => {
-    logger.error(err);
+    logger.error('Error in request', JSON.stringify(err));
     res.status(404).send(err.message);
```
ryansmith94: One of the issues was that the error message contained the text that you put in the URL which allowed html to be injected onto the page. Can you please change the response Chris to "Dashboard not found"?

h-kanazawa:
```suggestion
    res.status(404).send('Dashboard not found');
```

PR #1403, opened 2019-06-28
Update statementCount after statementsDeletion

h-kanazawa:

Closes https://learningpool.atlassian.net/browse/LL-54


lib/models/lrs.js
```diff
@@ -49,14 +49,12 @@ schema.plugin(auditRemove, {
   auditName: 'LRSAudit'
 });
 
-schema.statics.updateStatementCount = (lrs, cb) => {
+schema.statics.updateStatementCount = async (lrs) => {
```
ryansmith94: Nice work converting this to async.

worker/src/handlers/batchStatementDeletion/batchStatementDeletion.js
```diff
@@ -28,7 +29,7 @@ const markDone = async (batchDeleteId, jobDone) => {
 
 export default async ({
   batchDeleteId,
-  publish = pubishToQueue
+  publish = publishToQueue
```
ryansmith94: Thanks for correcting the typo.

api/src/routes/HttpRoutes.js
```diff
@@ -324,6 +326,14 @@ restify.serve(router, Statement, {
     return;
   },
   preUpdate: (req, res) => res.sendStatus(405),
+  postDelete: (req, _, next) => {
```
ryansmith94: For consistency going forward, do you think we should make this an async function? We could wrap the contents in a try-catch block to continue calling next with errors. Then we could await updateStatementCountsInOrg and call next after.

PR #1407, opened 2019-07-05
LL-48 viz refactor CustomColumnChart

h-kanazawa:
### Which issue does this close?

Closes https://learningpool.atlassian.net/browse/LL-48


ui/src/containers/Visualisations/CustomBarChart/Editor.js
```diff
@@ -0,0 +1,165 @@
...
+  const [tabIndex, setTabIndex] = useState(0);
+
+  return (
+    <div className="row">
+      <div className="col-md-6 left-border">
+        <div>
+          <DescriptionForm
+            visualisationId={model.get('_id')}
+            description={model.get('description', '')}
+            onChange={(description) => {
```
h-kanazawa: We can optimise more by *using `useMemo` hooks*
```
onChangeDescription = useMemo((description) => {
  updateModel({
    schema: 'visualisation',
    id: model.get('_id'),
    path: 'description',
    value: description,
  });
}, [model.get('_id')]);

...

<DescriptionForm
  ...
  onChange={onChangeDescription}
  />
```

or *moving the code defining this handler to ui/src/containers/Visualisations/CustomBarChart/index.js
(Define handler in `compose`)

I'd like to hear others' opinions

ui/src/containers/Visualisations/CustomBarChart/index.js
```diff
@@ -0,0 +1,25 @@
...
+import { connect } from 'react-redux';
+import { updateModel } from 'ui/redux/modules/models';
+import { setInMetadata } from 'ui/redux/modules/metadata';
+import Editor from './Editor';
+
+/**
+ * @param {immutable.Map} props.model - visualisation model
+ * @param {string} props.orgTimezone
+ */
+const CustomBarChard = compose(
```
ryansmith94: Just a little typo here.
```suggestion
const CustomBarChart = compose(
```

ui/src/containers/Visualisations/CustomBarChart/index.js
```diff
@@ -0,0 +1,25 @@
...
+  connect(
+    () => ({}),
+    { updateModel, setInMetadata },
+  ),
+)(Editor);
+
+CustomBarChard.propTypes = {
```
ryansmith94:
```suggestion
CustomBarChart.propTypes = {
```

ui/src/containers/Visualisations/CustomBarChart/index.js
```diff
@@ -0,0 +1,25 @@
...
+  model: PropTypes.instanceOf(Map).isRequired,
+  orgTimezone: PropTypes.string.isRequired,
+};
+
+export default CustomBarChard;
```
ryansmith94:
```suggestion
export default CustomBarChart;
```

ui/src/containers/Visualisations/components/index.js
```diff
@@ -0,0 +1,19 @@
+import AddQueryButton from './AddQueryButton';
```
ryansmith94: I probably wouldn't use index files if possible, I'd rather just import the components directly to reduce a bit of code maintenance. Not sure what you think?

ui/src/containers/Visualisations/index.js
```diff
@@ -0,0 +1,6 @@
+import CustomBarChart from './CustomBarChart';
```
ryansmith94: Again same as previous comment about index files.

ui/src/containers/Visualisations/components/index.js
h-kanazawa: I think importing from index file is simpler for developers.

The current CustomBarChart/Editor.js is like below

```
import {
  AddQueryButton,
  BarChartGroupingLimitForm,
  DescriptionForm,
  PreviewPeriodPicker,
  SourceViewForm,
  StackedSwitch,
  TimezoneForm,
} from '../components';
```

Other chart files will import many components from '../components' directory.

If we delete the index file and import from each component files, CustomBarChart/Editor.js will be

```
import AddQueryButton from '../components/AddQueryButton';
import BarChartGroupingLimitForm from '../components/BarChartGroupingLimitForm';
import DescriptionForm from '../components/DescriptionForm';
import PreviewPeriodPicker from '../components/PreviewPeriodPicker';
import SourceViewForm from '../components/SourceViewForm';
import StackedSwitch from '../components/StackedSwitch';
import TimezoneForm from '../components/TimezoneForm';
```

ryansmith94: Yeah definitely shorter and simpler to read. Just considering the experience for adding, removing, and renaming components.

When you add a component you might not be aware of the need to change the index file so we lose some consistency (although we have code reviews to hopefully catch that).

When you remove the component, again you need to know to delete the component file and export in the index, you also have to potentially check for references via the index file and the component file just in case someone forgets to import it from the index.

When you rename the component, you have to rename it in the component file, then find all the references to it via the index file and the component file. You also have to rename it in the index file too.

I think it's okay, not too much extra maintenance effort, just something we'll have to be weary of in code reviews to keep the consistency.

h-kanazawa: OK. I follow the style.

h-kanazawa:
This PR changes Bar Chart instead of Column Chart

ryansmith94:
Ah okay no worries, thanks @h-kanazawa :+1:

PR #1418, opened 2019-08-01
feat(Worker): Allows subset of workers to reduce processing (LL-128 & LLC-9)

h-kanazawa:
https://learningpool.atlassian.net/browse/LLC-9
https://learningpool.atlassian.net/browse/LL-128


worker/src/handlers/statement/allowedWorkerQueues.js
```diff
@@ -0,0 +1,51 @@
...
+  if (allowedWorkerQueuesString === '') {
+    return new Map();
+  }
+
+  return allowedWorkerQueuesString.split(',').reduce(
+    (accMap, queueString) => {
+      if (allowableWorkerQueues.has(queueString)) {
+        return accMap.set(queueString, queueString);
+      }
+      logger.warn(`"${queueString}" is ignored as an allowed worker queue. Allowable worker queues are "${STATEMENT_QUERYBUILDERCACHE_QUEUE}", "${STATEMENT_EXTRACT_PERSONAS_QUEUE}", and "${STATEMENT_FORWARDING_QUEUE}"`);
```
ryansmith94: I like the warning here, good idea.

worker/src/handlers/statement/allowedWorkerQueues.js
```diff
@@ -0,0 +1,51 @@
...
+  STATEMENT_EXTRACT_PERSONAS_QUEUE,
+  STATEMENT_FORWARDING_QUEUE,
+} from 'lib/constants/statements';
+
+/**
+ * allowable Worker Queues
+ *
+ * @type {Map<string, true>}
+ */
+const allowableWorkerQueues = new Map([
```
ryansmith94: Could this just be an array of these constants instead of a map to make it simpler?

worker/src/handlers/statement/allowedWorkerQueues.js
ryansmith94: Do you think it might be better to map over the `allowableWorkerQueues` to get the queue names and join them by commas to produce the last part of this warning rather than duplicating the list of allowable worker queues for easier maintenance?

worker/src/handlers/statement/allowedWorkerQueues.js
h-kanazawa: Yes it can be Array. Sorry, this implement was coded before I got to know the allowable queues is just 3. Array.includes is O(N), but Map.has is O(1). Additionally, this function is called every time statement comes. So I used Map for the performance. But size of allowable queues is 3, Array.includes is fast as well. I'll rewrite with array.

worker/src/handlers/statement/allowedWorkerQueues.js
h-kanazawa: OK. I just wanted to join the last 2 words with ", and".

ryansmith94: Ah yeah, I hadn't spotted the "and", happy for us to just display commas though if it's easier (not looked at the changes since this comment just yet).

worker/src/handlers/statement/allowedWorkerQueues.js
ryansmith94: Yeah appreciate the O(N) vs O(1) what you did makes sense, I don't think the array will get too large. Okay 👍

worker/src/handlers/statement/allowedWorkerQueues.js
```diff
@@ -0,0 +1,51 @@
...
+const getAllowedWorkerQueues = (allowedWorkerQueuesString) => {
+  if (allowedWorkerQueuesString === undefined) {
+    return allowableWorkerQueues;
+  }
+
+  if (allowedWorkerQueuesString === '') {
+    return [];
+  }
+
+  return allowedWorkerQueuesString.split(',').reduce(
```
ryansmith94: We could technically use `filter` instead of `reduce` here, but I'm not too worried.

h-kanazawa: Exactly! I'll rewrite here with filter.

ryansmith94: Okay thanks Hota 👍

PR #1438, opened 2019-08-16
feat(Visualisations): Adds options to display rows for min, max, average, and total for visualisations shown as tables. (LL-151)

h-kanazawa:

Closes 
- https://learningpool.atlassian.net/browse/LL-151
- https://learningpool.atlassian.net/browse/LL-45
- https://learningpool.atlassian.net/browse/LL-46
- https://learningpool.atlassian.net/browse/LL-47
- https://learningpool.atlassian.net/browse/LL-49

- Enable show stats in SourceResults
- Viz Refactor custom template PIE, COUNTER, XVSY, and FREQUENCY


lib/models/visualisation.js
```diff
@@ -90,6 +90,8 @@ const schema = new mongoose.Schema({
   statementContents: [statementContents],
   hasBeenMigrated: { type: Boolean, default: false },
   timezone: { type: String },
+  showStatsAtTop: { type: Boolean, default: true },
+  showStatsAtBottom: { type: Boolean, default: true },
```
lydiamross: Is it correct that by default, stats are shown at both top and bottom instead of either/or ? i.e. perhaps only one of the initialisers should be true.

h-kanazawa: @lydiamros thank you for reviewing.

@Ian247 Are stats displayed at bottom or top, or both of them?

ui/src/containers/VisualiseResults/SourceResults.js
```diff
@@ -141,50 +209,75 @@ const SourceResult = ({
     return <NoData />;
   }
 
+  const showStatsAtTop = visualisation.get('showStatsAtTop', true);
+  const showStatsAtBottom = visualisation.get('showStatsAtBottom', true);
+
+  const stats = calcStats(formattedResults);
+
+  // TODO: Show a message telling that the result might be limited
```
cbishopvelti: Can we do this?

h-kanazawa: @Ian247 @ryansmith94 When the size of results from API server is 10,000, the result may be limited. That means that `total`, `avg`, `min` and `row count` are probably not accurate numbers. Should we display a message if the result size is 10,000? If yes, can you specify the message text?

ryansmith94: Yeah I think we should display a message, something like "Totals calculated from the first 10,000 records".

Ian247: Are we saying that we could only total up 10,000 statements?

ryansmith94: No just 10,000 rows.

lib/models/visualisation.js
Ian247: It should be a toggle in the options to show stats at the top or bottom.

h-kanazawa: As a result of discussed with @Ian247, we decided to put `show totals` toggle and `top / bottom` toggle. So stats at top or bottom is exclusive.

h-kanazawa:
> The ticket says that the totals should be fixed when scrolling to keep them visible but the rest of the rows would disappear as you scroll through the table. Does this not work in reality?

I discussed with @ryansmith94, fixing totals became out of scope of this PR because the design and UX issue. However `ScrollableTable` component has the options to fix header and footer. Fixing is easy work.

lydiamross:
@Ian247 see Hota's reply a couple of days ago about the scrolling:

> I discussed with @ryansmith94, fixing totals became out of scope of this PR because the design and UX issue. However `ScrollableTable` component has the options to fix header and footer. Fixing is easy work.

ryansmith94:
@lydiamross good spot, don't worry though. Discussed it with Ian and decided that now we don't offer the option to display top and bottom at the same time they can stick. Hota originally implemented this with the stickiness but when top and bottom were shown together you didn't get to see any data rows because there wasn't enough room.

h-kanazawa:
@Ian247 I capitalised the label "bottom / top". As we discussed, sticking table head/foot is still OFF.

ryansmith94:
Thanks @h-kanazawa please add the stickiness before this is code reviewed again (see [my earlier comment](https://github.com/LearningLocker/learninglocker/pull/1438#issuecomment-528735917) about why stickiness is ok now)

h-kanazawa:
@ryansmith94 Thanks. I read your comment.
This morning, I discussed stickiness with Ian and Ben. Our decision is no stickiness at the moment because sticked tables collapse the design when dashboard widget height is short.

ryansmith94:
Ah okay cool, thanks @h-kanazawa :+1:

PR #1441, opened 2019-09-02
feat(Visualisations): New aggregation API with web sockets. (LL-139)

cbishopvelti:

Closes https://learningpool.atlassian.net/browse/LL-139.

Adds the initialise aggregation processor api.


api/src/controllers/AggregationProcessorController-test.js
```diff
@@ -0,0 +1,29 @@
+import { expect } from 'chai';
```
ryansmith94: I'm sure this probably doesn't match what we've done elsewhere in the controllers folder, but I think we should make a aggregationProcessor subfolder with these tests and the request handlers contained within that.

worker/src/handlers/aggregationProcessor/aggregationProcessor.js
```diff
@@ -0,0 +1,222 @@
...
+    ...(JSON.parse(model.pipelineString))
+  ];
+
+  return subtractPipeline;
+};
+
+const aggregationProcessor = async ({
+  aggregationProcessorId,
+  publishQueue = publish,
+}, done) => {
```
ryansmith94: I don't think we need a `done` callback since this is async. I think we should avoid having callbacks where possible.

ryansmith94:
Superseded by #1470

PR #1448, opened 2019-09-06
fix(Roles): Resolves issue for old roles where the "All" scope was selected for Org Admins but did not allow them to manage user's roles. (LL-169)

h-kanazawa:
Closes https://learningpool.atlassian.net/browse/LL-169


api/src/controllers/UserOrganisationSettingsController.js
```diff
@@ -43,7 +43,7 @@ const validateUpdatableKeys = (body, scopes) => {
 
   const updatingKeys = Object.keys(body);
 
-  if (scopes.includes(MANAGE_ALL_USERS)) {
+  if (scopes.some(s => [MANAGE_ALL_USERS, ALL].includes(s))) {
```
h-kanazawa: - Allow ALL scope to update 'filter', 'roles', and 'scopes' in the user organisation settings.

lib/services/auth/selects/models/user.js
```diff
@@ -2,7 +2,7 @@ import intersection from 'lodash/intersection';
 import getScopesFromAuthInfo
   from 'lib/services/auth/authInfoSelectors/getScopesFromAuthInfo';
 import { ALL, SITE_ADMIN } from 'lib/constants/scopes';
-import { MANAGE_ALL_USERS, ALL as ORG_ALL } from 'lib/constants/orgScopes';
```
h-kanazawa: `ALL` in orgScopes.js is undefined

PR #1454, opened 2019-09-19
refactor(Visualisations): Replaces VisualiseFilterForm for easier maintenance. (LL-190)

h-kanazawa:

https://learningpool.atlassian.net/browse/LL-190

- Replace `VisualiseFilterForm` with `FiltersForm`
  - `VisualiseFilterForm` has `switch (visualisation.type)` statements. We'd like to minimize `switch (visualisation.type)` in our project. Instead, we can define `FilterForm` in each Visualisation/*/Editor.js.


ui/src/containers/VisualiseFilterForm/index.js
```diff
@@ -1,176 +0,0 @@
...
-  onChangeTimezone = (value) => {
-    this.props.updateModel({
-      schema: 'visualisation',
-      id: this.props.model.get('_id'),
-      path: ['timezone'],
-      value
-    });
-  }
-
-  renderTabbedQueryBuilder = (labelled = true, defaults = {}, text = 'Build your query') => {
```
h-kanazawa: I dug how `defaults` is used. I found it is not used. So replacing code does not have `defaults`

ui/src/containers/VisualiseFilterForm/index.js
```diff
@@ -1,176 +0,0 @@
...
-  renderFrequency = () => this.renderTabbedQueryBuilder();
-
-  renderDefault = () => this.renderTabbedQueryBuilder();
-
-  render = () => {
-    const { model } = this.props;
-    const type = model.get('type');
-    const queries = model.get('filters', new List());
-
-    switch (type) {
```
h-kanazawa: Main purpose of this PR is deleting this switch statement

ui/src/containers/Visualisations/components/FilterForm.js
```diff
@@ -69,6 +69,7 @@ const FilterForm = ({
       <div className="form-group">
         <ColorPicker
           color={filter.get('color', VISUALISATION_COLORS[index])}
+          onChangeHashToForceRender={filter.hashCode()}
```
ryansmith94: Essentially there seem to be some checks in one of the HOCs to prevent re-renders when non-function props aren't changed. Since the color will be the same after the query is changed, it meant that component was not re-rendered and therefore the onChange function still held a reference to the old filter which contained the old query. When the color was changed, this meant that the function updated the entire filter with the old query.

PR #1459, opened 2019-09-20
refactor: Removes Cypress and tidies up associated files. (LL-193)

ryansmith94:
https://learningpool.atlassian.net/browse/LL-193


.eslintrc
```diff
@@ -78,7 +124,6 @@
     "__GOOGLE_ENABLED__": true,
     "socket": true,
     "webpackIsomorphicTools": true,
-    "cy": true,
-    "Cypress": true
+    "cy": true
```
lydiamross: @ryansmith94 Is it right that line 127 is still here?

ryansmith94: I think it actually can go, but I wasn't 100% so I've left it there for now.

PR #1476, opened 2019-11-11
Renamed Curatr to Headstream[LL-217]

cbishopvelti:

Closes https://learningpool.atlassian.net/browse/LL-217.


ui/src/containers/Visualisations/TemplateStreamActivitiesWithMostComments/buildModel.js
```diff
@@ -18,6 +18,7 @@ const filter = fromJS({
         $comment: '{"criterionLabel":"B","criteriaPath":["statement","context","platform"]}',
         'statement.context.platform': {
           $in: [
+            /* TODO: Should it be changed too? */
```
crazy-grizzly: @ryansmith94 , could you please take a look on this?

ryansmith94: Ooosh that's a good question, tagging in James (@ht2) because he'll be able to find that out.

ht2: Great question! I think we add "Stream" as an additional param in the $in, but in fact we haven't updated that yet and will come as part of a bigger change later in the year when we also need to change the domain of curatr3.com.... More on that to follow, but for the mo I suggest...
```suggestion
            "Stream",
```

crazy-grizzly: @ht2 , got it. Thank you 👍

PR #1478, opened 2019-11-19
fix(Login Page): Upgrades passport plugin to fix Google logins. (LL-275)

### LINKED ISSUE #1477: Google Authentication Not Working
#### What version were you using?
v4.10.2 initially

#### What steps can we follow to reproduce the behaviour?
Attempt a sign-in with Google on an existing or new Learning Locker setup. In my case, the deployment was setup in August 2019 and was working correctly. Sometime in October login stopped working with no change made to the setup.

#### What is the actual behaviour?

Sign in does not work.

API instance logs this error which is caused by the `passport-google-oauth2` v1.0.0 making a call for user profile data to Google+ API's that have or are being shutdown. Going to the link in the Cloud console throws an error and doesn't allow for enabling the Legacy People API in new projects.

https://github.com/jaredhanson/passport-google-oauth2/blob/v1.0.0/lib/strategy.js#L54

```
InternalOAuthError: failed to fetch user profile (status: 403 data: {
  "error": {
    "code": 403,
    "message": "Legacy People API has not been used in project <project ID> before or it is disabled. Enable it by visiting https://console.developers.google.com/apis/api/legacypeople.googleapis.com/overview?project=<project ID> then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry.",
    "status": "PERMISSION_DENIED",
    "details": [
      {
        "@type": "type.googleapis.com/google.rpc.Help",
        "links": [
          {
            "description": "Google developers console API activation",
            "url": "https://console.developers.google.com/apis/api/legacypeople.googleapis.com/overview?project=<project ID>"
          }
        ]
      }
    ]
  }
})
``` 

Error also occurs here (https://github.com/LearningLocker/learninglocker/blob/master/api/src/auth/passport.js#L215) since an email won't be successfully found in the `profile`:

```
TypeError: Cannot read property 'value' of undefined
at Strategy._verify (/app/api/dist/server/webpack:/auth/passport.js:215:46)
at /app/node_modules/passport-oauth2/lib/strategy.js:202:24
at /app/node_modules/passport-google-oauth20/lib/strategy.js:122:5
at passBackControl (/app/node_modules/oauth/lib/oauth2.js:134:9)
```

#### What is the expected behavior?
Google sign-in should successfully login the user.

#### Is there any additional information that will help us replicate/understand the problem?

v2.0.0 of `passport-google-oauth` was created to handle this: https://github.com/jaredhanson/passport-google-oauth2/pull/51

Corresponding blog post for reference: https://medium.com/passportjs/google-api-shutdown-330c3b47e3df

I will open a PR for this issue as well.

davidstoker:
https://learningpool.atlassian.net/browse/LL-275
Closes #1477 

Upgrade `passport-google-oauth` to v2.0.0 to use version that no longer uses the deprecated Google+ APIs for login. The `type` property on a profile email is not returned by Passport. Instead, key off the only other property of `verified`. This was manually tested when investigating the root cause of the initial issue.


ryansmith94:
Hey @davidstoker thanks for reporting this issue and submitting the PR, I'm reviewing this now, apologies for the delayed response.

ryansmith94:
Hey @davidstoker, I can't seem to get this to pass our CircleCI build so I can't merge this right now, could you please take a look at some point?

davidstoker:
@ryansmith94 Are you able to re-run CircleCI without caching? Looks like a bcrypt problem which is not being changed here. This person indicated problems with bcrypt and caching on CircleCI (https://discuss.circleci.com/t/npm-bcrypt-binary-not-saved-to-cache/7773)

You'll notice that the first run did execute tests but failed due to  "serviceBus should subscribe to queue: Error: timeout of 5000ms exceeded." which also wouldn't be directly related to the changes being made here

ryansmith94:
Rerunning it now @davidstoker. We may need to make the same change I've made in another PR https://github.com/LearningLocker/learninglocker/pull/1493/files#diff-1d37e48f9ceff6d8030570cd36286a61R32

ryansmith94:
Strange that the service bus is timing out every time I run this now, even though it's fine on other branches. I'll try updating this branch to merge in the latest changes from master.

ryansmith94:
Ok annoyingly I couldn't figure out what was causing an issue here, so I've created a new PR and that seems to work. Circle doesn't let me build without cache anymore for whatever reason so the cache was probably messed up for this branch. See #1501.

davidstoker:
Thanks @ryansmith94! Re-deployed our setup today and happy to have Google auth working again for users

ryansmith94:
That's good to hear @davidstoker 👍 Thanks for your patience.

PR #1486, opened 2019-12-10
fix(Personas): Disables add attribute button when name and value are empty (LL-219)

lydiamross:

[LL-219](https://learningpool.atlassian.net/browse/LL-219)


lydiamross:
> Nice, looks okay to me. No need to change anything, just wanted you to know that if this was TypeScript it would have forced you to use `foo.length !== 0 && bar.length !== 0` because of falsy values (it's a long story, you can look it up - https://www.sitepoint.com/javascript-double-negation-trick-trouble/).

Yeah thanks, I considered the longhand but given that it works in this case, I thought it was easier to read

PR #1499, opened 2020-01-09
Persona Imports error on Field type selection[LL-284]

crazy-grizzly:
Closes [LL-284](https://learningpool.atlassian.net/browse/)


crazy-grizzly:
Btw, code completion now works really good now 🤔 

Screenshot 2020-01-14 at 10 09 08
Screenshot 2020-01-14 at 10 08 49
Screenshot 2020-01-14 at 10 10 08

crazy-grizzly:
Also I noticed that Webstorm's code analysis system goes crazy using next code:

```
/** @typedef {import("./personasImport.types").PersonasImportStructure} PersonasImportStructure */
```
So the CPU is under load all the time because code analysis is looped between type exported from `.ts` file and JSDoc's notation in `.js` file. As the solution different names of types can be used:

`.ts.` file:

```export interface ISuperDataStructure```
or
```export interface SuperDataStructureInterface```

`.js` file:

```/** @typedef {import("./personasImport.types").ISuperDataStructure} SuperDataStructure */```
or
```/** @typedef {import("./personasImport.types").SuperDataStructureInterface} SuperDataStructure */```

File: lib/models/personasImport.types.ts [hunk 1]
```diff
@@ -0,0 +1,47 @@
...
+  constant?: string;
+}
+
+export interface PersonasImportStructure {
+  [key: string]: PersonaImportField;
+}
+
+export interface PersonasImport {
+  title: string;
+  owner: any; // TODO: define type
```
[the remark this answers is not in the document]
crazy-grizzly: Yeah, I know. But these changes are for the discussion now. If we decide to use this, I'll add related data types

.eslintignore
```diff
@@ -1,4 +1,5 @@
 webpack/*
 karma.conf.js
 tests.webpack.js
-dist
\ No newline at end of file
+dist
+/**/*.d.ts
```
ryansmith94: Is there any reason we're ignoring these definition files and have used the `.d.ts` extension instead of `.types.ts`?

lib/connections/mongoose.js
```diff
@@ -3,6 +3,8 @@ import Promise from 'bluebird';
 import logger from 'lib/logger';
 import defaultTo from 'lodash/defaultTo';
 
+/** @typedef {import("./mongoose.types").Connection} Connection */
```
ryansmith94: Is there any reason we can't do the following instead to avoid having our own file to mirror the Mongoose types?

```
/** @typedef {import('mongoose').Connection} Connection */
```

lib/connections/mongoose.js
```diff
@@ -26,9 +29,15 @@ const createConnection = () => {
   });
 };
 
+/** @type {Object.<string, Connection>} */
```
ryansmith94: I don't think you meant to put the dot here before the chevron.

lib/models/personasImport.types.ts
```diff
@@ -0,0 +1,47 @@
+import { Document, Model, Schema } from 'mongoose';
```
ryansmith94: Thanks for putting these types close to the code that uses them, I think that'll make them easier to maintain, I do worry that without full TS support and compilation checks these may get out of date and cause confusion, what do you think Vadim?

.eslintignore
crazy-grizzly: Sorry, this is from previous tries and should be removed

lib/connections/mongoose.js
crazy-grizzly: I thought that it will not work because trying to jump to that type shows dropdown with corresponding type selection
Screenshot 2020-01-16 at 18 31 24
Screenshot 2020-01-16 at 18 31 27

But it seems it really works:
Screenshot 2020-01-16 at 18 32 50

But in this case we have an issue that I described in comment to this PR related to CPU load

lib/connections/mongoose.js
crazy-grizzly: True 🙂

lib/models/personasImport.types.ts
crazy-grizzly: Yeah, I agree with you. This is not a good way to do so because of redundant files. I though that it works only so but tried approach with `module:mongoose.XXX` and it works too. So I removed those `types.ts.` files. Please take a look on latest changes

ryansmith94: Ok thanks Vadim.

.eslintignore
ryansmith94: Ah okay no worries.

lib/connections/mongoose.js
ryansmith94: Ok so you're having performance issues with this?

crazy-grizzly: Now everything is ok, but as the solution we need to use different name for `@typedef` type like so:

```
/** @typedef {module:mongoose.Connection} MongooseConnection */
```

PR #1511, opened 2020-01-28
References to organisations still exist after deletion (LL-311 & LLC-26)

PrinceWaune:
https://learningpool.atlassian.net/browse/LLC-26
https://learningpool.atlassian.net/browse/LL-311


lib/models/organisation.js
```diff
@@ -217,98 +217,108 @@ schema.post('init', function handleInit() {
...
 
-      // check if the user has settings for this org
-      const existingOS = find(existingOrganisationSettings, setting =>
-         setting.organisation.toString() === organisation._id.toString()
-      );
+  const existingOS = find(existingOrganisationSettings, setting =>
+    setting.organisation.toString() === organisation._id.toString()
+  );
+
+  if (existingOS) {
```
PrinceWaune: In which case the existingOS variable will be true? As far as I understand the current orgSetting not exist in the owner yet. Maybe we only need to push the current orgSetting to the owner.orgSettings?

lib/models/organisation.js
```diff
@@ -217,98 +217,108 @@ schema.post('init', function handleInit() {
...
-      title: 'Observer',
-      owner_id: organisation.owner,
-      organisation: organisation._id,
-      scopes: [
-        VIEW_PUBLIC_DASHBOARDS,
-        VIEW_PUBLIC_VISUALISATIONS,
-        VIEW_PUBLIC_JOURNEYS
-      ]
-    });
+  const createAdminRole = await Role.create({
```
PrinceWaune: Why are we creating roles before finding the user? What if we don't find the owner by organisation.owner.

PrinceWaune: Why are we creating 2 roles? Chris answered me: Just by default. Sorry, I don't understand. 
Is a subset of observer role accesses a part of accesses of administrator role, is not it?

lib/models/organisation.js
```diff
@@ -217,98 +217,108 @@ schema.post('init', function handleInit() {
 schema.pre('save', async function preSave(next) {
   const User = getConnection().model('User');
   const Role = getConnection().model('Role');
+  const organisation = this;
+  const objectId = mongoose.Types.ObjectId;
 
   // https://github.com/Automattic/mongoose/issues/1474
   this.wasNew = this.isNew;
```
PrinceWaune: In our case, we have no postSave hook. Why are we using this trick?

lib/models/organisation-test.js
```diff
@@ -2,10 +2,71 @@ import chai, { expect } from 'chai';
 import chaiAsPromised from 'chai-as-promised';
 import { Error } from 'mongoose';
 import Organisation from './organisation';
+import User from './user';
+import Role from './role';
+import { createOrganisation, createUser } from './test-utils/testModelsHelper';
```
crazy-grizzly: In case when we will refactor other places or cover another models with tests, we need to create a helper class with util methods for tests. But for now I guess it's ok

lib/models/test-utils/testModelsHelper.js
```diff
@@ -0,0 +1,10 @@
+import User from '../user';
+import { userFixture } from './fixtures/user.fixtures';
+import Organisation from '../organisation';
+import { orgFixture } from './fixtures/organisation.fixtures';
+
+export const createUser = async () =>
```
crazy-grizzly: Here you should pass `user` as an argument to this method. It's bad idea to simply use predefined fixture by default

lib/models/test-utils/testModelsHelper.js
```diff
@@ -0,0 +1,10 @@
...
+  await new User(userFixture).save();
+
+export const createOrganisation = async () =>
```
crazy-grizzly: Yeah, better to pass org fixture here as an argument too

lib/models/organisation.js
```diff
@@ -217,98 +217,108 @@ schema.post('init', function handleInit() {
 schema.pre('save', async function preSave(next) {
   const User = getConnection().model('User');
   const Role = getConnection().model('Role');
+  const organisation = this;
+  const objectId = mongoose.Types.ObjectId;
```
crazy-grizzly: Better to save letter case and also you can use desctructuring assignment like so: 

```
const { ObjectId } = mongoose.Types;
```

lib/models/organisation-test.js
```diff
@@ -2,10 +2,71 @@ import chai, { expect } from 'chai';
...
+    await Role.deleteMany({});
+    await Organisation.deleteMany({});
+    await User.deleteMany({});
+  });
+
+  it('should properly create organisation', async () => {
+    const user = await createUser();
+    const organisation = await createOrganisation();
+
+    const userAfterCreateOrganisation = await User.findOne({
```
crazy-grizzly: I guess it will look better like so: `userAfterOrganisationCreated`

lib/models/organisation-test.js
```diff
@@ -2,10 +2,71 @@ import chai, { expect } from 'chai';
...
+      .find(orgSetting =>
+        orgSetting.organisation.toString() === organisation._id.toString()
+      );
+    const createdUsersRoles = await Role.find({
+      organisation: organisation._id
+    });
+
+    await expect(createdUsersOrganisation).to.not.be.undefined;
+    await expect(createdUsersOrganisationSetting).to.not.be.undefined;
+    await expect(createdUsersRoles).to.be.an('array').to.have.lengthOf(2);
```
crazy-grizzly: I guess it would be nice also to check those roles for proper structure and names, etc.

crazy-grizzly: But wait, in this case it should be placed into `User`'s model tests, so nevermind

lib/models/organisation.js
PrinceWaune: I deleted it because it is not necessary

lib/models/test-utils/testModelsHelper.js
PrinceWaune: Done!

lib/models/test-utils/testModelsHelper.js
PrinceWaune: Done!

lib/models/organisation-test.js
PrinceWaune: Done!

lib/models/organisation.js
ryansmith94: We create 2 roles by default one called "Admin" which has access to the Settings and one called "Observer" that does not have access to the Settings. We have done this by default as a convenience for our users. Previously when we didn't have the roles feature we had two types of user "Admin" and "Observer" so the naming is due to legacy to maintain backwards compatibility in our user experience.

ryansmith94: We want these two roles to exist regardless of whether we can find the owner. This prevent issues caused by concurrency where the owner is deleted while we're creating the organisation.

lib/models/organisation.js
ryansmith94: Yeah if we have no post save then it appears this is no longer necessary, it's possible we did have a post save hook in the past. Happy for you to remove this line if you can't find references to this `wasNew` property anywhere else in the codebase.

ryansmith94: Please comment on this thread with what you decide (if you decide to remove or keep this line) then resolve this comment.

ryansmith94: Actually I can see this being used a bit further down, but it's used within this function so we could just use `this.isNew` in its place.

lib/models/organisation.js
ryansmith94: Hmmmm thanks for flagging this, I'm actually not sure how we'd ever reach this code, I can only guess that it's there just in case the situation occurs somehow, perhaps while restoring an organisation that was deleted.

lib/models/organisation.js
PrinceWaune: I think we need to delete this line of code for better readability. If we need to add a postSave hook in the future, just return this line of code.

lib/models/organisation.js
PrinceWaune: Okay, I understand. Thank you very much for your answer!

lib/models/organisation.js
PrinceWaune: I think this "if" statement never runs because as far as I can understand we have a bug inside "if". I will try to explain what I mean. Suppose the "updatedSettings" variable
```
  const existingOrganisationSettings = owner.organisationSettings || [];
  let updatedSettings = existingOrganisationSettings;
```
contains such an array
```
[organisationSettingObject1, organisationSettingObject2, organisationSettingObject3]
```
When we dive inside our "if" statement 
```
 if (existingOS) {
    const updatedOS = existingOS;

    updatedOS.roles = updatedOS.roles
      ? union(updatedOS.roles, newRoles)
      : updatedOS.roles = newRoles;

    updatedSettings = map(updatedSettings, (os) => {
      if (os.organisation.toString() === organisation._id.toString()) {
        return updatedOS;
      }
    });
```
we map over the "updatedSettings" variable. Because inside a map function we have "if" statement
```
      if (os.organisation.toString() === organisation._id.toString()) {
        return updatedOS;
      }
```
 the map function returns to us such an array 
```
[OrganisationSettingObject n, undefined, undefined, undefined...]. 
```
After that, we just put this array to the owner.orgSettings 
```
 owner.organisationSettings = updatedSettings;
```
save the owner and overwrite owner.orgSettings with "undefined" values in our database.

So I suppose testers noticed that while creating, updating, restoring organization data in owner overwritten

PrinceWaune: I apologize if this is difficult to understand

ryansmith94: Yeah well spotted! No it wasn't difficult to understand, thanks for pointing that out and explaining it as clearly as you could. Since no one has reported this issue, I believe that is a decent enough indicator that this code is never reached, in which case let's remove this if block.

ryansmith94: Removing this if block will also get rid of the warning from the LGTM bot.
Screenshot 2020-02-17 at 08 22 00

ryansmith94: Once you've removed this block @PrinceWaune please move this ticket to Ready for QA as it's already been reviewed by nearly everyone on the team at this point.

PrinceWaune: Yeah, great!

PrinceWaune:
https://github.com/LearningLocker/learninglocker/blob/bc8a2b23139bc05844a3d31cd886f0fb24c35732/lib/models/organisation.js#L341-L345

Can an organization really have multiple owners?

PrinceWaune:
We set the ownerOrganization in only two places
https://github.com/LearningLocker/learninglocker/blob/bc8a2b23139bc05844a3d31cd886f0fb24c35732/lib/models/user.js#L270-L272
and
https://github.com/LearningLocker/learninglocker/blob/bc8a2b23139bc05844a3d31cd886f0fb24c35732/lib/models/user.js#L372-L374
the code shows that this only happened when we are creating the user.
Accordingly, based on this, no matter how many new organizations we will create, according to the predicate 
```
ownerOrganisation: this._id
```
https://github.com/LearningLocker/learninglocker/blob/bc8a2b23139bc05844a3d31cd886f0fb24c35732/lib/models/organisation.js#L341-L345
, records will never be found and, accordingly, nothing will be updated.
ownerOrganisationSettings will be updated only in case of updating the organization that was associated with the user during creation

PrinceWaune:
When do we create a new organization maybe we will need to push this organization to ownerOrganisation field in a user document?
Actual behavior: 
When creating a new organization fields ownerOrg and ownerOrgSettings in a user document remain the same due to the logic that was described in the previous comment.

ryansmith94:
> Can an organization have multiple owners?
If so, @ryansmith94 please can you tell me a case so that I can reproduce it?
> [learninglocker/lib/models/organisation.js](https://github.com/LearningLocker/learninglocker/blob/bc8a2b23139bc05844a3d31cd886f0fb24c35732/lib/models/organisation.js#L341-L345)

Theoretically I suppose it could via the database, but in practice I think it can only have one if it's created via regular methods in the UI or API.

ryansmith94:
> the code shows that this only happened when we are creating the user.
> Accordingly, based on this, no matter how many new organizations we will create, according to the predicate

@PrinceWaune yeah that's correct, the owner organisation is the organisation that the user was first invited by and we use this to determine the password rules for the user.

lib/models/organisation.js
```diff
@@ -210,108 +258,103 @@ schema.statics.limitIdsByOrg = function limitIdsByOrg(
...
+    this.expirationNotifications.weekBeforeNotificationSent = EMAIL_NOOP;
+  }
 
-    await Promise.all([createAdminRole, createObserverRole]);
+  if (!isUndefined(this._oldTimezone) && this.timezone !== this._oldTimezone) {
+    await update$dteTimezoneInDB(this._id, this.timezone);
   }
 
-  // Update users settings
+  // TODO: I think the organisationSettings are never updated.
```
lydiamross: By this TODO, do you mean that the following updateMany on the user doesn't work?
I think this line can be removed or reverted

PrinceWaune: TODO was removed

Issue #1440, opened 2019-08-28
fix(Site Admin > Users): Fails to load organisations for users that are members of lots of organisations. (LL-178)


kampernet:
using v4.4.0

**Steps to reproduce**
1. fresh install
2. create over 200 orgs 
3. go to /admin/users as site admin
4. open your user for editing

**What happens:**
The URL generated ( the API call to filter orgs ) passes in every org id, url encoded.  This becomes massively long (over 11k characters) and fails.  Either the browser refuses to send it or the server can't handle it.

<img width="1651" alt="Screen Shot 2019-08-28 at 11 18 31 AM" src="https://user-images.githubusercontent.com/2286714/63874085-71968000-c986-11e9-8358-2c52dfbb0d31.png">

**Expected behaviour:** 
Pagination or some other method of scaling to over 200 orgs.

kampernet:
this is one of the problem lines: 

https://github.com/LearningLocker/learninglocker/blob/master/ui/src/containers/Home/index.js#L250

if I can change that to make a call more like http://localhost:3000/api/v2/user/<authUser._id>?populate=organisations and extract the orgs somehow, that would fix it.

I've never devved in a node app like this before though, so I've not been able to figure out yet how to do that with all the libs at play

ryansmith94:
Populate would be okay but there's no pagination within the population with Express Restify Mongoose. I think we need to create a new API that allows you to use POST requests with these filters in the body. Perhaps we could look into a GraphQL implementation instead.

kampernet:
Yeah I couldn't find a way to do it that felt "correct".  I don't know enough about MERN or GraphQL to comment intelligently.  It looked like it'd require some kind of change to ConnectionController.js to support populate.  

At any rate, for our purposes, in my fork, it worked to just splice the first 100.  On the containers/SiteUsers/SiteUsersOrgItems.js it doesn't do pagination anyways and only shows the first 10.  

On the front page, I changed it so that if the number of userOrgs was > 50 I show the search, and if the search is populated then I use the searchFilter, otherwise just show the first 100... instead of merging the filters.

ryansmith94:
Tracked internally by https://learningpool.atlassian.net/browse/LL-178

PR #1513, opened 2020-01-30
502 errors on site users page when user is a member of lots of organisations (LL-307 & LLC-34)

PrinceWaune:
https://learningpool.atlassian.net/browse/LLC-34
https://learningpool.atlassian.net/browse/LL-307

Closes https://github.com/LearningLocker/learninglocker/issues/1440


ui/src/pages/SiteUsersPage/SiteUserOrgItems.js
```diff
@@ -11,25 +11,34 @@ const enhance = compose(
...
+      {
+        hasMore
+          ? <button
+            className={'btn btn-default btn-sm'}
+            onClick={fetchMore}>
+            Load more
+          </button>
+          : null
+      }
+    </div>
```
vladislav-999: It can be replaced by a simplified construction
`{
        hasMore
          && <button
            className={'btn btn-default btn-sm'}
            onClick={fetchMore}>
            Load more
          </button>
      }`

crazy-grizzly: Personally I prefer to move conditional blocks to separate methods and naming them. This can improve code readability. Sure this should be applied to blocks of code larger than one simple tag or something

PrinceWaune: Done!

PrinceWaune:
@ryansmith94 I have a question about the correct work of functional. Can you help me, please?
So, what problem we had? When we try to display all organizations of a user we build request like this (for example)
```
API/connection/organisation=filter
```
where filter was
```
{
  '_id': {
    'in': user.organisations
  }
}
```
and user.organisations was
```
[orgId1, orgId2, orgId3, ..., orgIdN]
```
If we have a lot of organization, we have trouble with a very very long query string.
Since we have a many-to-many relation between organizations and users collections, we can fix the problem as follows.
Build request like this (for example)
```
API/connection/organisation=filter
```
where filter was
```
{
  'users['id']': user._id
}
```
but in our case, we don't have users field in an "Organization document" like organizations field in a "User document".
We can change schema in our model file and add saving users in a hook method while saving an organization model. But organizations that are already stored in the database will not have any associated users.
So, maybe we need to create a migration for this case or we must build another request on the frontend side?

ryansmith94:
@PrinceWaune yeah you've understood the problem correctly. Ideally we would have a OrganisationUser model with a `userId` and `orgId` on it to link users. However that's quite a significant change at the moment, so what I proposed was two solutions in the meantime. The easy solution which is to display the first 10 organisations, then say "plus X more not displayed". The harder solution is to display the first 10 organisations, but allow users to load the next 10, and then the next 10, etc (like we do on the Data > Source page for statements). In future we should absolutely have this extra OrganisationUser model.

PrinceWaune:
Yes, a great solution! But, I will try to explain why we can't realize the harder solution correct. In our case, we can't realize a harder solution because we don't have OrganisationUser model. Therefore, all that remains for us build request like this
```
API/connection/organisation=filter
```
where filter was
```
{
  '_id': {
    'in': user.organisations.slice(0, 10);
  }
}
```
**BUT** because we have a problem which is described in https://learningpool.atlassian.net/browse/LL-311 in response we can get not 10 but a random count of organizations (from 10 to 0). So I think, this behavior will not be clear to the user. 
@ryansmith94 What do you think about this?

ryansmith94:
I think this solution where it shows between 0 and 10 organisations is acceptable for now given that it's better than the user receiving a strange error. That issue will be resolved once LL-311 is merged which should be soon anyway.

PrinceWaune:
> There are a lot of TODO in this code. Not sure I'm fan of that, I understand there are future tickets to better resolve this issue, but probably better to explain why the code is currently the way is its, rather than leaving TODOs around.

Yes I described it here

https://github.com/LearningLocker/learninglocker/blob/b1beb67d70eba276de868b0be01b4580b3e195bb/ui/src/pages/SiteUsersPage/SiteUserOrgItems.js#L28-L29

lib/models/plugins/addCRUDFunctions.js
```diff
@@ -171,6 +171,7 @@ async function getConnection({
 
   const direction = (before) ? BACKWARDS : FORWARDS;
 
+  // TODO: Blocks of code in if and elseif statements are equal.
```
ryansmith94: That is interesting, @cbishopvelti do you think there is a typo in this code? Looks like the `else if` block below should have set the cursor to `before` instead of `after`.

ryansmith94: In fact, I'm going to say it's definitely a typo, can you please change that @PrinceWaune and remove this TODO?

ryansmith94: Looks like the else if block below should have set the cursor to before instead of after, can you please change that @PrinceWaune and remove this TODO?

ui/src/pages/SiteUsersPage/SiteUserOrgItems.js
```diff
@@ -11,12 +11,11 @@ const enhance = compose(
   }),
   withProps(({ user }) => {
     const schema = 'organisation';
-    const organisations = user
-      .get('organisations')
-      .map(org => new Map({ $oid: org }));
+    const organisations = user.get('organisations').map(org => new Map({ $oid: org }));
+    // TODO: https://github.com/LearningLocker/learninglocker/pull/1513#issuecomment-587064642
```
ryansmith94: I don't think this is a todo, just a useful link, so let's remove the TODO part of this comment and change it to "Decision made via"

lib/models/plugins/addCRUDFunctions.js
ryansmith94: > This is core functionality. I don’t think it’s a good idea to edit this here

I agree, in this case, since it's a small change, let's just do it here to save us some time and management.

ryansmith94: We don't paginate backwards in our UI anywhere so this isn't something QA will have tested previously.

PrinceWaune: > Looks like the else if block below should have set the cursor to before instead of after, can you please change that @PrinceWaune and remove this TODO?

Yeah, sure!

ui/src/pages/SiteUsersPage/SiteUserOrgItems.js
```diff
@@ -25,11 +25,37 @@ const enhance = compose(
...
+    })
     .valueSeq();
-  return <ul>{orgsItems}</ul>;
+
+  const countOfRemainingOrganisations = user.get('organisations').count() - orgItems.count();
+
+  return (
+    <div>
+      <ul>{orgItems}</ul>
+      {countOfRemainingOrganisations > 0 ? <p>Plus { countOfRemainingOrganisations } more not displayed</p> : null}
```
vladislav-999: It can be reduced
`{countOfRemainingOrganisations > 0 && <p>Plus { countOfRemainingOrganisations } more not displayed</p>}`

PrinceWaune: Done!

PR #1515, opened 2020-02-05
fix(Statement Deletion): Fixes store recounts when clients have no LRS attached (LL-322)

PrinceWaune:
https://learningpool.atlassian.net/browse/LL-322.

### DIFF
--- a/lib/models/lrs.js
+++ b/lib/models/lrs.js
@@ -9,16 +9,16 @@ import * as scopes from 'lib/constants/scopes';
 import addCRUDFunctions from 'lib/models/plugins/addCRUDFunctions';
 import auditRemove from 'lib/models/plugins/auditRemove';
 
-const createDefaultClient = async function () {
-  if (!this.title) {
-    this.title = 'New xAPI store';
+const createDefaultClient = async (lrsModel) => {
+  if (!lrsModel.title) {
+    lrsModel.title = 'New xAPI store';
   }
   const ClientModel = getConnection().model('Client');
   const client = new ClientModel({
-    organisation: this.organisation,
-    lrs_id: this._id,
+    organisation: lrsModel.organisation,
+    lrs_id: lrsModel._id,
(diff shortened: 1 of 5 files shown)


crazy-grizzly:
@PrinceWaune , could you please annotate code inside `lrs.js` like it done in `personasImport.js` using JSDoc?

PrinceWaune:
> @PrinceWaune , could you please annotate code inside `lrs.js` like it done in `personasImport.js` using JSDoc?

PrinceWaune:
> @PrinceWaune , could you please annotate code inside `lrs.js` like it done in `personasImport.js` using JSDoc?

Sure, no problem!

lib/models/lrs.js
```diff
@@ -9,6 +9,40 @@ import * as scopes from 'lib/constants/scopes';
...
+ *  @property {number} statementCount
+ */
+
+/** @typedef {module:mongoose.Model<Lrs>} lrsModel */
+
+/**
+ * @param {lrsModel} lrsModel
+ * @returns {Promise<void>}
+ */
+const createDefaultClient = async (lrsModel) => {
```
ryansmith94: Good idea extracting this function, makes the code easier to understand.

lib/models/lrs.js
```diff
@@ -58,10 +83,14 @@ schema.statics.updateStatementCount = async (lrs) => {
   });
 };
 
-schema.statics.decrementStatementCount = async (lrs) => {
+/**
```
ryansmith94: I think we need to change the way this function is called in "lib/services/lrs.js" since it seems to be passing the full model, although we might be able to delete the `decrementStatementCount` function from that file since I don't think it's used anymore.

PrinceWaune: Oh, cool! I didn’t even notice it, thanks!

PrinceWaune: Done!

PR #1517, opened 2020-02-11
fix(Google Login): Prevents uninvited users gaining access (LLC-10)

crazy-grizzly:
https://learningpool.atlassian.net/browse/LLC-10
https://learningpool.atlassian.net/browse/LL-295

### DIFF
--- a/ui/src/utils/oauth.js
+++ b/ui/src/utils/oauth.js
@@ -78,7 +78,6 @@ function getEndpoint(provider) {
 }
 
 export function openPopup(provider) {
-  console.log('openPopup', provider);
   return window.open(getEndpoint(provider), provider, `${settings},${getPopupDimensions(provider)}`);
 }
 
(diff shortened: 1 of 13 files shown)


ui/src/utils/oauth.js
```diff
@@ -78,29 +78,80 @@ function getEndpoint(provider) {
 }
 
 export function openPopup(provider) {
+  console.log('openPopup', provider);
```
ryansmith94: I think we should probably remove this.

ui/src/utils/oauth.js
```diff
@@ -78,29 +78,80 @@ function getEndpoint(provider) {
...
+
+/**
+ * @param {Window} popup
+ * @param {checkForToken~tokenResolve} resolve
+ * @param {checkForToken~tokenReject} reject
+ */
 function checkForToken(popup, resolve, reject) {
-  if (popup.closed) reject({ errors: 'Authentication was cancelled.' });
-  else {
+  console.log('checkForToken');
```
ryansmith94: Again should probably remove this.

crazy-grizzly: Ah damn 🙂 Maybe we should setup eslint rule for debug messages?

ryansmith94: Yeah agreed.

Issue #1288, opened 2018-12-19
Mongo migration error (LL-132)


caperneoignis:
You need to replace the parts of this template where you see {{some text}}.

**Version**
master at  "Revision caf7683b63bf32d9b2288aa3b37f3ee23c0761d5"  Comment: "chore(deps): update dependency @ht2-labs/semantic-release to v1.1.50"  on 2.6.1 

**Steps to reproduce the bug**
1. update to newest LL version using deploy.sh script
2. at end of install run mongo migration

**Expected behaviour**
no error

**Actual behaviour**
````
Error: Db migration (20180212_expiration) does not match (20180320093000_shareable_dashboards)
    at /usr/local/learninglocker/releases/ll-20181219-7894ed0792e96941a3193ab057e35f9ba083a5b7/webapp/cli/dist/server/webpack:/migrateMongo/index.js:97:13
    at List.__iterate (/usr/local/learninglocker/releases/ll-20181219-7894ed0792e96941a3193ab057e35f9ba083a5b7/webapp/node_modules/immutable/dist/immutable.js:2206:13)
    at List.forEach (/usr/local/learninglocker/releases/ll-20181219-7894ed0792e96941a3193ab057e35f9ba083a5b7/webapp/node_modules/immutable/dist/immutable.js:4381:19)
    at _callee2$ (/usr/local/learninglocker/releases/ll-20181219-7894ed0792e96941a3193ab057e35f9ba083a5b7/webapp/cli/dist/server/webpack:/migrateMongo/index.js:92:17)
    at tryCatch (/usr/local/learninglocker/releases/ll-20181219-7894ed0792e96941a3193ab057e35f9ba083a5b7/webapp/node_modules/regenerator-runtime/runtime.js:65:40)
    at GeneratorFunctionPrototype.invoke [as _invoke] (/usr/local/learninglocker/releases/ll-20181219-7894ed0792e96941a3193ab057e35f9ba083a5b7/webapp/node_modules/regenerator-runtime/runtime.js:303:22)
    at GeneratorFunctionPrototype.prototype.(anonymous function) [as next] (/usr/local/learninglocker/releases/ll-20181219-7894ed0792e96941a3193ab057e35f9ba083a5b7/webapp/node_modules/regenerator-runtime/runtime.js:117:21)
[...]

PR #1524, opened 2020-03-03
feat(CLI): Adds command to the order of migrations (LL-132 & LLC-11)

cbishopvelti:
https://learningpool.atlassian.net/browse/LLC-11
https://learningpool.atlassian.net/browse/LL-132
https://github.com/LearningLocker/learninglocker/issues/1288

This command can be run with `node cli/dist/server migrateMongo --fixOrder`


cli/src/migrateMongo/index.js
```diff
@@ -80,8 +81,35 @@ const getOutstandingMigrations = async ({ down, up, migrations }) => {
...
+    toSave = [dbMigration, ...toSave];
+
+    inc += 1;
+  }
+
+  await bmap(toSave, (toSav) => {
+    const out = toSav.save();
+    return out;
+  });
+};
```
crazy-grizzly: How about so? 🤔 

```suggestion
/**
 * @param {OrderedMap} migrations
 * @returns {Promise<void>}
 */
export const fixDbMigrationOrder = async ({ migrations }) => {
  const dbMigrations = await migration.find({}).sort({ order: 1 });
  const dbMigrationsToSave = [];

  migrations
    .keySeq()
    .toJS()
    .forEach(
      (migrationItemKey) => {
        const dbMigration = find(
          dbMigrations,
          dbMigrationItem => dbMigrationItem.key === migrationItemKey
        );

        if (!dbMigration) {
          return;
        }

        dbMigration.order = dbMigrationsToSave.length + 1;
        dbMigrationsToSave.push(dbMigration);
      }
    );

  await bmap(dbMigrationsToSave, dbMigrationToSave => dbMigrationToSave.save());
};
```

PR #1536, opened 2020-04-21
fix(139): negative numbers in aggregation

cbishopvelti:

Closes issue on 139, where statements can be inserted in the middle of the window.

Fixes statements being inserted into the middle of the window.


.env.example
```diff
@@ -269,3 +269,6 @@ FS_REPO=local
 
 # JWT_ACCESS_TOKEN_EXPIRATION=1h
 # JWT_REFRESH_TOKEN_EXPIRATION=7d
+
+# AGGREGATIEN BLOCK SIZE, default 1 days
```
ryansmith94, line 273:
```suggestion
# AGGREGATION BLOCK SIZE, default 1 days
```

worker/src/handlers/aggregationProcessor/aggregationProcessor-test.js
```diff
@@ -261,4 +265,112 @@ describe('aggregationProcessor', () => {
     expect(done).to.equal(true);
     expect(result[0].count).to.equal(1);
   });
+
+  describe('after the fact statements', () => {
```
ryansmith94, line 269: Tests here look clear, nice job 👍

worker/src/handlers/aggregationProcessor/aggregationProcessor.js
```diff
@@ -214,12 +237,34 @@ const getAddPipeline = async ({
...
+            $gte: moment(model.fromTimestamp || now).toDate(),
+            $lt: getAddFromTimestamp({ model, now }).toDate()
+          }
+        },
+        {
+          stored: {
+            $gte: moment(model.lastRun).toDate(),
+          }
+        }
+      ]
```
ryansmith94, line 257: This should be functionally equivalent.
```suggestion
          timestamp: {
            $gte: moment(model.fromTimestamp || now).toDate(),
            $lt: getAddFromTimestamp({ model, now }).toDate()
          },
          stored: {
            $gte: moment(model.lastRun).toDate(),
          }
```

worker/src/handlers/aggregationProcessor/aggregationProcessor.js
```diff
@@ -214,12 +237,34 @@ const getAddPipeline = async ({
     };
   }
 
+  // MIDDLE
+
+  let addToMiddle;
+  if (model.lastRun) {
```
ryansmith94, line 243: Probably easier to use a ternary here instead of the ternary you have later in the code as shown below.

```js
...(addToMiddle ? [addToMiddle] : []),
```
So we can change the code here to the following, replacing the `...` with the code below these lines.
```
const addToMiddle = model.lastRun === undefined ? [] : [{ /* ... */ }]
```
Then we can remove the ternary later, I've added a code suggestion for that.

worker/src/handlers/aggregationProcessor/aggregationProcessor.js
```diff
@@ -214,12 +237,34 @@ const getAddPipeline = async ({
...
+  }
+
   return [
     {
       $match: {
         $or: [
           { timestamp: addToFrontPipeline },
-          { timestamp: addToEnd }
+          { timestamp: addToEnd },
+          ...(addToMiddle ? [addToMiddle] : []),
```
ryansmith94, line 267: See earlier comment about moving the ternary earlier.
```suggestion
          ...addToMiddle,
```

worker/src/handlers/aggregationProcessor/aggregationProcessor.js
```diff
@@ -275,7 +310,8 @@ const aggregationProcessor = async (
   {
     aggregationProcessorId,
     publishQueue = publish,
-    now // For testing
+    now, // For testing
+    actualNow // for testing
```
ryansmith94, line 314: Can we give these two variables clearer names? Perhaps rename `now` to `endDate` and `actualNow` to `now` or something similar?

worker/src/handlers/aggregationProcessor/aggregationProcessor.js
```diff
@@ -365,7 +403,17 @@ const aggregationProcessor = async (
...
+      lastRun: actualNow.toDate(),
+      lastCompletedRun: hasReachedEnd({
+        model: {
+          ...model,
+          toTimestamp: toTimestamp.toDate(),
+          fromTimestamp: fromTimestamp.toDate()
+        },
+        now
+      }) ?
+        actualNow.toDate() : model.lastCompletedRun
```
ryansmith94, line 416: I think this code would be easier to read if the result of `hasReachedEnd` was assigned to a variable above the `newModel` definition.

worker/src/handlers/aggregationProcessor/aggregationProcessor.js
cbishopvelti: Yes, I didn't really come up with better names. Now is where the window is calculated from, so the case where this wouldn't be the same as actualNow is if benchmarking is enabled. Then 'now' would be actualNow - windowSize.

But I want to give it some obscure long name, as now is the most import variable in the algorithm, where as actualNow is just to catch the edge case of future insertions.

PR #1541, opened 2020-06-23
fix(deps): Updates Mongo to 4.2 (LLC-234)

lydiamross:
Closes [LLC-234](https://learningpool.atlassian.net/browse/LLC-234)

#### Breaking Changes for Merge Commit
BREAKING CHANGE: Removes `maxScan` parameter from API
BREAKING CHANGE: Requires Mongo 4.2

### DIFF
--- a/cli/src/commands/v2-migrations/20171122100800_common_indexes.js
+++ b/cli/src/commands/v2-migrations/20171122100800_common_indexes.js
@@ -29,10 +29,9 @@ const createStatementIndexes = (connection) => {
   createIndexWithVoid('verbId_objId', { ...verbIdIndex, ...objectIdIndex });
   createIndexWithVoid('mbox', { 'statement.actor.mbox': 1 });
   createIndexWithVoid('account', accountIndex);
-  createIndexWithVoid('timestamp__id', { timestamp: -1, _id: -1 });
   createIndexWithVoid('stored__id', { stored: -1, _id: -1 });
   createIndex(stmts, 'statementId_lrs_id', { 'statement.id': 1, lrs_id: 1 });
-  createIndexWithOrg('timestamp__id', { timestamp: -1, _id: 1 });
+  createIndexWithOrg('timestamp__id', { timestamp: -1, _id: -1 });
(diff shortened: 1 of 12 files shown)


cli/src/commands/v2-migrations/20171122100800_common_indexes.js
```diff
@@ -29,10 +29,9 @@ const createStatementIndexes = (connection) => {
   createIndexWithVoid('verbId_objId', { ...verbIdIndex, ...objectIdIndex });
   createIndexWithVoid('mbox', { 'statement.actor.mbox': 1 });
   createIndexWithVoid('account', accountIndex);
-  createIndexWithVoid('timestamp__id', { timestamp: -1, _id: -1 });
   createIndexWithVoid('stored__id', { stored: -1, _id: -1 });
   createIndex(stmts, 'statementId_lrs_id', { 'statement.id': 1, lrs_id: 1 });
-  createIndexWithOrg('timestamp__id', { timestamp: -1, _id: 1 });
+  createIndexWithOrg('timestamp__id', { timestamp: -1, _id: -1 });
```
lydiamross, line 34: Added an amalgamation of the two - it seems to ignore the order when deciding if it's a duplicate

lydiamross, line 34: Added an amalgamation of the two - it seems to ignore the order when deciding if it's a duplicate

ryansmith94: Thanks @lydiamross, @ht2 you'll definitely want to take a look at this.

ryansmith94:
Added breaking changes for merge commit to PR description.

ctm8788:
This change seems like it will prevent any users who deploy to Azure and use CosmosDB mongoAPI from being able to deploy the application. Currently the API is only at 3.6

ryansmith94:
@ctm8788 that is very true, we appreciate your concern, we've never officially supported CosmosDB and always recommend using Mongo.

PR #1561, opened 2020-10-04
Update "@google-cloud/pubsub": to version 0.19.1

SwampApe:
Couldn't compile on Ubuntu 20.04 unless version was bumped up because prebuilt binaries weren't available for gRPC. Version 0.19.1 is the latest version without breaking changes.


SwampApe:
Any ideas on why azure service bus breaks when pubsub is updated? It seems like sbService fails for some reason on line 21.

